PR #5, opened 2015-09-16
Update to latest sbt-houserules, formatting

Duhemm:


dwijnand:
:+1:

eed3si9n:
The new houserules has many of the Bintray settings, so we should clean them up from build.sbt as well.

Duhemm:
@eed3si9n Done

dwijnand:
Also the licenses, developers and org can be dropped:
- https://github.com/sbt/sbt-houserules/blob/v0.2.1/src/main/scala/sbt/houserules/BintraySettingPlugin.scala#L17-L25
- https://github.com/sbt/sbt-houserules/blob/v0.2.1/src/main/scala/sbt/houserules/BuildPlugin.scala#L14

Duhemm:
You're right, sorry. I just removed them.

dwijnand:
Nice :+1:

PR #6, opened 2015-09-17
add toString to UpdateOptions

jozic:
as of 0.13.9 there is no an easy way to see updateOptions attributes

'[info]     sbt.UpdateOptions@f5a7a399'

### DIFF
--- a/librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
+++ b/librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
@@ -1,6 +1,5 @@
 package sbt.librarymanagement
 
-import java.io.File
 import org.apache.ivy.plugins.resolver.DependencyResolver
 import org.apache.ivy.core.settings.IvySettings
 import sbt.util.Logger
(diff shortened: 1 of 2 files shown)


eed3si9n:
This class will likely be be generated by sbt-datatype eventually, and it'll regress even if I merge this.
@Duhemm maybe we need to generate `toString`?

Duhemm:
@eed3si9n Current sbt-datatype already generates `toString` for both Java and Scala

jozic:
cool, so if this going to be fixed in next version, feel free to close this PR

jvican:
Is this still an issue? @Duhemm says that `toString` is already generated by sbt-datatype.

jozic:
that's what i see now with 0.13.15

```sbt
> updateOptions
[info] sbt.UpdateOptions@bc690cbc
> sbtVersion
[info] 0.13.15
> 
```

Duhemm:
@jozic We don't use sbt-datatype / contraband in 0.13.

Also, it doesn't look like `UpdateOptions` is generated in 1.0? I guess it's because of `UpdateOptions.ResolverConverter`.

dwijnand:
Along with finishing the remove sbinary/pickling we should finish migrating from case classes to contraband classes.

jvican:
Is that only concerning this module or all sbt modules? @dwijnand

dwijnand:
All of them.

dwijnand:
I don't see any issue merging this, particularly given it has a test. If/when it gets ported to by contraband generated, we can tweak the test that its toString is intelligible.

eed3si9n:
This merged uncompilable code.

PR #9, opened 2015-10-02
Split RichUpdateReport into internal

eed3si9n:
The main change is d6cadf79171875798f3d9c92d52352f6d76526bf.
This is so we can use the injected methods on `UpdateReport` from the scripted tests. See https://github.com/sbt/sbt/pull/2226#issuecomment-144661630

/review @Duhemm, @dwijnand, @jsuereth


deltag.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+for t in `git tag`
+do
+  git tag -d $t
+done
+
```
dwijnand: woops?

eed3si9n: oops

dwijnand:
LGTM pending Travis

Duhemm:
Was this already included in 0.1.0-M4? Any reason why it has not been merged?

eed3si9n:
0.1.0-M4 is c97563a0c0df4c0c35421ed33e78495ab6e50dc4.
I'll merge this now.

PR #11, opened 2015-11-11
Don't warn on inconsistent versions in different configurations

Duhemm:
(This is a port of sbt/sbt#2258)

sbt was reporting warning abouts inconsistent versions of dependencies
even if these dependencies didn't have the same configuration (as in
`provided` vs `compile`).

This commit fixes this problem by comparing the dependencies by
organization, artifact name and configuration.


dwijnand:
:+1:

PR #14, opened 2015-12-10
make internal configuration private and add .gitignore

He-Pin:


He-Pin:
we have `internalMap`,maybe I don't have a full context currently,just following the book,is it directly referenced anywhere?,currently I could directly reference to it in build.sbt. eg. `Configurations.CompileInternal`

eed3si9n:
It's not directly referenced, but they leak out via `ivyConfigurations`, so I don't know if it's a good idea to completely hide them. If it's `private[sbt]` then it gives the internal code chance to fix things when needed.

PR #16, opened 2016-01-16
Travis JDKs

dwijnand:


dwijnand:
Split in two commits, because I expect it Travis to fail in EvictionWarningSpec under Java 8 like it does for me locally...

eed3si9n:
Could you fix the test so it accepts both patterns?

```
[info] - should print out message about the eviction if it's enabled *** FAILED ***
[info]   List("There may be incompatibilities among your library dependencies.", "Here are some of the libraries that were evicted:", " * commons-io:commons-io:1.4 -> 2.4 (caller: net.databinder:unfiltered-uploads_2.10:0.8.0, ca.gobits.bnf:bnfparser:1.0)") was not equal to List("There may be incompatibilities among your library dependencies.", "Here are some of the libraries that were evicted:", "    * commons-io:commons-io:1.4 -> 2.4 (caller: ca.gobits.bnf:bnfparser:1.0, net.databinder:unfiltered-uploads_2.10:0.8.0)") (EvictionWarningSpec.scala:166)
```

dwijnand:
Yep

PR #20, opened 2016-01-17
FPORT: Refix snapshots

dwijnand:
Foward port of sbt/sbt#2313.


ivy/src/main/scala/sbt/CompatibilityWarning.scala
```diff
@@ -0,0 +1,26 @@
+package sbt
```
eed3si9n, line 1: I merged it already but is this going to a wrong path?

dwijnand, line 1: Yeah it is, sorry, I'll move it.

PR #21, opened 2016-01-17
Move CompatibilityWarning to sbt.internal.librarymanagement

dwijnand:
Moved to sbt.internal.librarymanagement, let me know if that's wrong.


eed3si9n:
```
[error] /home/travis/build/sbt/librarymanagement/librarymanagement/src/main/scala/sbt/internal/librarymanagement/CompatibilityWarning.scala:4: not found: type Logger
[error]   def apply(module: IvySbt#Module, mavenStyle: Boolean, log: Logger): Unit = {
[error]                                                              ^
[error] 
```

PR #26, opened 2016-02-10
Add Ivy `branch` support to ModuleID 

oxbowlakes:
Allow an Ivy branch to be set on the project's module (such that Ivy is aware of it), as well as on its dependencies.
This change:
- adds the `branchName: Option[String]` to `ModuleID`. This change is not indended to be backwards-compatible.
- changes the resolution cache's artifact pattern to include the `([branch]/)` after the module name
- changes `defaultIvyPatterns` as well as the local ivy repository pattern to include the `([branch]/)` after the module name
- changes `ModuleReport` generation to set the branch onto the `ModuleID` (on top of saving it inside the `ModuleReport`)

Afterwards, a later pull request to [sbt 1.0.x](https://github.com/sbt/sbt/tree/1.0.x) will:
- fix `sbt.CacheIvy` to include `branchName` in `Format[ModuleID]`, and add a `CacheIvyTest` that ensures a cached-then-read `ModuleID` remains equal to the original (save for `CrossVersion`, which is not being cached perfectly).

Usage:

``` scala
libraryDependencies += "org.something" % "module" % "version" branch "myBranch"
```

or

``` scala
libraryDependencies += "org.something" % "module" % "version" branch Some("myBranch")
```

### DIFF
--- a/librarymanagement/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
+++ b/librarymanagement/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
@@ -422,7 +422,7 @@ private[sbt] object IvySbt {
   def toID(m: ModuleID) =
     {
       import m._
-      ModuleRevisionId.newInstance(organization, name, revision, javaMap(extraAttributes))
+      ModuleRevisionId.newInstance(organization, name, branchName.orNull, revision, javaMap(extraAttributes))
     }
 
   private def substituteCross(m: ModuleSettings): ModuleSettings =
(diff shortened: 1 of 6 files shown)


dansanduleac:
Note that https://github.com/sbt/sbt/pull/1994 contains the context explaining our motivation and use-case for this.
However, unlike the case for that PR, since 1.0 is not yet released, we don't aim to maintain binary compatibility here.

eed3si9n:
Since Ivy natively has the notion of [branch](http://ant.apache.org/ivy/history/2.3.0/terminology.html#branch), I think this is a good timing to finally pull this into sbt.
As @jsuereth noted, it might make our notion of dependency graph more Ivy than before, but it's currently ingrained into the code at most level, so I am not too concerned about that. If we are serious about making the backend swappable with coursier or Aether, and have the implementation, we can just throw error saying that the graph with branch is not supported.

LGTM from my end.

PR #29, opened 2016-03-02
Rename InlineConfigurationWithExcludes -> InlineConfiguration

dansanduleac:
This long-named class is no longer required, and it only exists so that `InlineConfiguration` could remain backwards-compatible with `0.13`, which is no longer necessary here.


dwijnand:
Great! LGTM

eed3si9n:
LGTM

PR #31, opened 2016-03-23
Fix small typo ("Sorrting" should be "Sorting")

JD557:
This PR fixes a small typo in `SbtChainResolver.scala` ("**Sorrting** results from $rmr, using ${rmr.getPublicationDate} and ${rmr.getDescriptor.getPublicationDate").

This typo is also on [sbt/sbt-zero-thirteen](https://github.com/sbt/sbt-zero-thirteen/blob/d9f10853fb5cece1c0e598d2dd08756a95f7bd43/ivy/src/main/scala/sbt/ivyint/SbtChainResolver.scala). Should I send another PR there?


dwijnand:
If anything you couldn't send it to https://github.com/sbt/sbt/tree/0.13 if you want. (sbt-zero-thirteen exists only for the convenience of github search).

LGTM

PR #46, opened 2016-06-21
Give IvyPaths a nicer toString

dwijnand:


Duhemm:
Will we generate `IvyPaths` with sbt-datatype? sbt-datatype generates an implementation of `toString`.

dwijnand:
Yeah, I assume `IvyPath` will be part of the API, so we'll migrate to sbt-datatype, but this is just a quick and easy fix.

PR #48, opened 2016-07-01
Make CrossVersion & ModuleId more commonly equal

dwijnand:
Two instance of functions defined in the same way don't equal themselves, but
the same instance of a function does. So by using a val idStringFun for full and
binary, and making Binary and Full case classes there are much better chances
that ModuleId's defined in the same way will be equal.

### DIFF
--- a/librarymanagement/src/main/scala/sbt/librarymanagement/CrossVersion.scala
+++ b/librarymanagement/src/main/scala/sbt/librarymanagement/CrossVersion.scala
@@ -26,6 +26,11 @@ object CrossVersion {
    */
   final class Binary(val remapVersion: String => String) extends CrossVersion {
     override def toString = "Binary"
+    override def hashCode = remapVersion.##
+    override def equals(that: Any) = that match {
+      case that: Binary => this.remapVersion == that.remapVersion
+      case _            => false
+    }
   }
 
   /**
@@ -35,6 +40,11 @@ object CrossVersion {
    */
   final class Full(val remapVersion: String => String) extends CrossVersion {
     override def toString = "Full"
+    override def hashCode = remapVersion.##
+    override def equals(that: Any) = that match {
+      case that: Full => this.remapVersion == that.remapVersion
+      case _          => false
+    }
   }
 
   private val disabledTag = implicitly[FastTypeTag[Disabled.type]]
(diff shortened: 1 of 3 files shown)


eed3si9n:
This reminds me I have to write a blog article.

dwijnand:
On?

eed3si9n:
On persisting things that we can't persist.

dwijnand:
Btw the workaround is assign CrossVersion.full to a val:

``` scala
private val crossVersionFull = CrossVersion.full // workaround sbt/librarymanagement#48
implicit final class ModuleIDWithCompilerPlugin(val _m: ModuleID) extends AnyVal {
  def compilerPlugin(): ModuleID = sbt.compilerPlugin(_m cross crossVersionFull)
}
```

then I can use `-=`:

``` scala
> show libraryDependencies
[info] List(org.scala-lang:scala-library:2.11.8, org.scalamacros:paradise:3.0.0-SNAPSHOT:plugin->default(compile), org.scalameta:scalameta:1.1.0-SNAPSHOT)
> set libraryDependencies -= "org.scalamacros" % "paradise" % "3.0.0-SNAPSHOT" compilerPlugin()
[info] Defining *:libraryDependencies
[info] The new value will be used by *:allDependencies, *:dependencyPositions
[info] Reapplying settings...
[info] Set current project to dataclasses (in build file:/d/dataclasses/)
> show libraryDependencies
[info] List(org.scala-lang:scala-library:2.11.8, org.scalameta:scalameta:1.1.0-SNAPSHOT)
```

dwijnand:
Is this good to go?

eed3si9n:
I don't think it is.. and here's the blog post - http://eed3si9n.com/registry-and-reference

dwijnand:
I think that blog post presents a good, general way to tackle persisting (and therefore equality).

But here I'm just trying to nudge the status quo of a couple of specific classes a little into improvement.

eed3si9n:
Sure but introducing case class on a public API is a bincompat liability that we eventually need to get rid of. http://www.scala-sbt.org/0.13/docs/Coding-Guideline.html#Avoid+case+classes%2C+use+sbt-datatype

dwijnand:
Doh! You're right. I can rework that.

dwijnand:
I implemented hashCode/equals to achieve the desired outcome.

eed3si9n:
LGTM

PR #49, opened 2016-07-14
Add Resolver.sbtIvyRepo

dwijnand:
.. deprecate & rename SbtPluginRepositoryRoot to SbtRepositoryRoot.

This is to address https://github.com/sbt/sbt/pull/2670#discussion_r70466677.

### DIFF
--- a/librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
+++ b/librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
@@ -174,7 +174,9 @@ object Resolver {
   private[sbt] def useSecureResolvers = sys.props.get("sbt.repository.secure") map { _.toLowerCase == "true" } getOrElse true
 
   val TypesafeRepositoryRoot = typesafeRepositoryRoot(useSecureResolvers)
-  val SbtPluginRepositoryRoot = sbtPluginRepositoryRoot(useSecureResolvers)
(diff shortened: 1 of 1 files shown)


librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
```diff
@@ -174,7 +174,9 @@ object Resolver {
   private[sbt] def useSecureResolvers = sys.props.get("sbt.repository.secure") map { _.toLowerCase == "true" } getOrElse true
 
   val TypesafeRepositoryRoot = typesafeRepositoryRoot(useSecureResolvers)
-  val SbtPluginRepositoryRoot = sbtPluginRepositoryRoot(useSecureResolvers)
+  val SbtRepositoryRoot = sbtPluginRepositoryRoot(useSecureResolvers)
```
eed3si9n: Should we rename `def sbtPluginRepositoryRoot(...)` to `def sbtRepositoryRoot(...)`

eed3si9n:
Minor comment, but generally LGTM

librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
dwijnand: Good catch. Pushed.

PR #58, opened 2016-11-16
Switch to sbt-datatype, sjson-new & Vectors

dwijnand:
Review please by @eed3si9n.


dwijnand:
Or let me know how you think I should break this down and/or simplify it to ease reviewing it.

librarymanagement/src/main/scala/sbt/librarymanagement/Artifact.scala
```diff
@@ -5,30 +5,47 @@ package sbt.librarymanagement
 
 import java.io.File
 import java.net.URL
-import sbt.serialization._
 
-final case class Artifact(name: String, `type`: String, extension: String, classifier: Option[String], configurations: Iterable[Configuration], url: Option[URL], extraAttributes: Map[String, String]) {
+abstract class ArtifactParent {
```
eed3si9n: Is this the mixin for the methods? If so the naming convention I think should be `ArtifactExtra`.
Sort of like `ProjectExtra` etc - https://github.com/sbt/sbt/blob/0.13/main/src/main/scala/sbt/Project.scala#L608

eed3si9n (changes requested):
Minor comment, but mostly looks good.

librarymanagement/src/main/scala/sbt/internal/librarymanagement/formats/Function1Format.scala
```diff
@@ -0,0 +1,8 @@
+package sbt.internal.librarymanagement.formats
+
+import sjsonnew._
+
+trait Function1Format { self: BasicJsonProtocol =>
+  implicit def Function1Format[T, U]: JsonFormat[T => U] =
+    project(MyCrazyReferences.referenced _, (ref: String) => MyCrazyReferences(ref, classOf[T => U]))
```
dwijnand: @eed3si9n How do you feel about this?

librarymanagement/src/main/scala/sbt/librarymanagement/Artifact.scala
```diff
@@ -5,30 +5,47 @@ package sbt.librarymanagement
...
 
-object Artifact {
-  def apply(name: String): Artifact = Artifact(name, DefaultType, DefaultExtension, None, Nil, None)
-  def apply(name: String, extra: Map[String, String]): Artifact = Artifact(name, DefaultType, DefaultExtension, None, Nil, None, extra)
-  def apply(name: String, classifier: String): Artifact = Artifact(name, DefaultType, DefaultExtension, Some(classifier), Nil, None)
-  def apply(name: String, `type`: String, extension: String): Artifact = Artifact(name, `type`, extension, None, Nil, None)
-  def apply(name: String, `type`: String, extension: String, classifier: String): Artifact = Artifact(name, `type`, extension, Some(classifier), Nil, None)
-  def apply(name: String, url: URL): Artifact = Artifact(name, extract(url, DefaultType), extract(url, DefaultExtension), None, Nil, Some(url))
-  def apply(name: String, `type`: String, extension: String, classifier: Option[String], configurations: Iterable[Configuration], url: Option[URL]): Artifact =
+abstract class ArtifactCompanion {
```
dwijnand: @eed3si9n What about the companion? Should it be `ArtifactExtra` too?

eed3si9n: `ArtifactFunctions`, maybe.

librarymanagement/src/main/scala/sbt/internal/librarymanagement/formats/Function1Format.scala
eed3si9n: The name looks crazy. Anything containing the function itself probably should be handled as reference pattern, not the function.

eed3si9n (approved):

Issue #60, opened 2016-12-02
Match errors due to changes in MavenRepository and MavenCache class hierarchy


dwijnand:
I'm getting these when running scripted in sbt:

```
[info] scala.MatchError: cache:publish-m2-local: /Users/dnw/.m2/repository (of class sbt.librarymanagement.MavenCache)
[info] 	at sbt.internal.librarymanagement.ConvertResolver$$anonfun$defaultConvert$1.applyOrElse(ConvertResolver.scala:117)
[info] 	at sbt.internal.librarymanagement.ConvertResolver$$anonfun$defaultConvert$1.applyOrElse(ConvertResolver.scala:115)
[info] 	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
[info] 	at sbt.internal.librarymanagement.ConvertResolver$.apply(ConvertResolver.scala:112)
[info] 	at sbt.internal.librarymanagement.IvySbt$$anonfun$mapResolvers$1$1.apply(Ivy.scala:279)
[info] 	at sbt.internal.librarymanagement.IvySbt$$anonfun$mapResolvers$1$1.apply(Ivy.scala:279)
[info] 	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
[info] 	at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
[info] 	at scala.collection.Iterator$class.foreach(Iterator.scala:893)
[info] 	at scala.collection.AbstractIterator.foreach(Iterator.scala:1336)
[info] 	at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
[info] 	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
[info] 	at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
[info] 	at scala.collection.AbstractTraversable.map(Traversable.scala:104)
[info] 	at sbt.internal.librarymanagement.IvySbt$.mapResolvers$1(Ivy.scala:279)
```

The reason is that in the current (85e7cf68a56a076ae46bda0c5b76116ce6e8ae6f) version of lm MavenCache is no longer a subclass of MavenRepository, which is what `defaultConvert` is expecting.

Here's the class hierarchy in [v0.13.13](https://github.com/sbt/sbt/blob/v0.13.13/ivy/src/main/scala/sbt/Resolver.scala#L34-L52):

```scala
[...]

PR #61, opened 2016-12-05
One MavenRepository to rule them all? Refs #60

dwijnand:
@eed3si9n is this what you had in mind?

### DIFF
--- a/librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
+++ b/librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
@@ -26,6 +26,11 @@ final class RawRepository(val resolver: DependencyResolver) extends Resolver(res
     }
 }
 
+abstract class MavenRepositoryFunctions {
+  def apply(name: String, root: String, localIfFile: Boolean = true): MavenRepository =
+    new MavenRepo(name, root, localIfFile)
+}
+
 abstract class PatternsFunctions {
   implicit def defaultPatterns: Patterns = Resolver.defaultPatterns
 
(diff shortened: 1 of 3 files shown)


eed3si9n:
Yea. Pretty much.

librarymanagement/src/main/scala/sbt/librarymanagement/Resolver.scala
```diff
@@ -258,8 +258,8 @@ abstract class ResolverFunctions {
       new File(sbt.io.Path.userHome, ".m2/repository")
   }
   // TODO - should this just be the *exact* same as mavenLocal?  probably...
-  def publishMavenLocal: MavenCache = new MavenCache("publish-m2-local", mavenLocalDir)
-  def mavenLocal: IMavenRepository = new MavenCache("Maven2 Local", mavenLocalDir)
+  def publishMavenLocal: MavenRepository = new MavenRepository("publish-m2-local", mavenLocalDir)
```
eed3si9n: We should try to use `apply` when possible to unify towards it.

eed3si9n (changes requested):
Minor comment.

dwijnand:
I don't like this modelling with rootFile being an Option..

dwijnand:
And `root: String` and `rootFile: Option[File]` both being properties for the underlying datatype.

eed3si9n (approved):

eed3si9n:
I think it's an acceptable compromise on keeping the older `MavenRepository`-like behavior and the feature.
The alternative would be to turn `MavenRepository` into an interface, and come up with a sensible name for non-cache values.

dwijnand:
I prefer the alternative then. Aka option 2 I gave in #60

eed3si9n:
ok. AFAIK the only people using `MavenCache` is https://github.com/sbt/sbt-maven-resolver/blob/503843421fdf18c89e1af387a45ba9e9ddf1f1d7/src/main/scala/sbt/MavenResolverConverter.scala, so I'm fine with either solution as long as `MavenRepository`'s `apply` is provided.

dwijnand:
That easy. Much easier than MavenCache's MatchError'ing all over the shop.

PR #75, opened 2017-03-18
sbt-contraband 0.3.0-M4

eed3si9n:


dwijnand:
for what, btw?

eed3si9n:
This is so we can do "monorepo".

PR #76, opened 2017-03-31
Allow for some customization in CrossVersion results

dwijnand:
The old way to customize CrossVersion results was a `String => String`
function called 'remapVersion', removed in
301ec787f2f17367064cda5eef3a9009ba66e21f.

That was removed because it's not possible to serialise Function1s, and we
want to serialise CrossVersion (and therefore ModuleID, etc) to be able to
transfer them in sbt server.

This commit reintroduces a less-powerful way to vary the results of apply
CrossVersion, but just providing the opportunity to define a suffix. From
looking at the users of CrossVersion (Scala.js' sbt-scalajs-plugin & Scala
Native's sbt-crossproject) this looks to be sufficient.

### DIFF
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -138,12 +138,16 @@
           "namespace": "sbt.librarymanagement",
           "target": "Scala",
           "doc": [
-            "Cross-versions a module using the result of appending `suffix` to the binary version.",
-            "For example, if `suffix = \"_foo\"` and the binary version is \"2.10\",",
-            "the module is cross-versioned with \"2.10_foo\"."
+            "Cross-versions a module using the result of",
+            "prepending `prefix` and appending `suffix` to the binary version.",
+            "For example, if `prefix = \"foo_\"` and `suffix = \"_bar\"` and the binary version is \"2.10\",",
+            "the module is cross-versioned with \"foo_2.10_bar\"."
           ],
           "type": "record",
-          "fields": [ { "name": "suffix", "type": "String", "default": "\"\"", "since": "0.0.1" } ]
+          "fields": [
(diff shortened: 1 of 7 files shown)


librarymanagement/src/main/contraband/librarymanagement.json
```diff
@@ -138,19 +138,20 @@
           "namespace": "sbt.librarymanagement",
           "target": "Scala",
           "doc": [
-            "Cross-versions a module using the result of applying `remapVersion` to the binary version.",
-            "For example, if `remapVersion = v => \"2.10\"` and the binary version is \"2.9.2\" or \"2.10\",",
-            "the module is cross-versioned with \"2.10\"."
+            "Cross-versions a module using the result of appending `suffix` to the binary version.",
+            "For example, if `suffix = \"_foo\"` and the binary version is \"2.10\",",
+            "the module is cross-versioned with \"2.10_foo\"."
```
sjrd: Oh. In Scala.js and sbt-crossproject it's the other way around: `"foo_2.10"` (hence the full artifact name is `artifactname_foo_2.10`. This was not an arbitrary choice. We needed `_2.10` to remain the last part for tooling not to choke. For example, the "suspicious binary incompatibility" detector of the Scala IDE (which severely warns if you use a _2.10 dependency in a 2.11 project) requires the Scala binary version suffix to be last.

sjrd (changes requested):

librarymanagement/src/main/contraband/librarymanagement.json
dwijnand: You're right. The whole thing is suffixed onto the artifactId (after a `_`). Of course you need the ability to prefix that suffix. I'll change it. Thanks for looking.

sjrd:
Thanks :)

eed3si9n (approved):

PR #78, opened 2017-03-31
Expose RichUpdateReport & DependencyFilter

### LINKED ISSUE #42: Why are RichUpdateReport operations internal?
(Some of) the operations of `RichUpdateReport` are very useful, [publicly documented](http://www.scala-sbt.org/0.13/docs/Update-Report.html) methods. Why did they become internal?

We use `select` in several places in Scala.js. For example at https://github.com/scala-js/scala-js/blob/608baf718e4366dfdcb0745f13facb2b438344f6/sbt-plugin/src/main/scala/scala/scalajs/sbtplugin/ScalaJSJUnitPlugin.scala#L31
What are we supposed to use instead?

dwijnand:
Fixes #42


librarymanagement/src/main/scala/sbt/librarymanagement/RichUpdateReport.scala
```diff
@@ -17,13 +15,17 @@ final class RichUpdateReport(report: UpdateReport) {
...
+  private[sbt] def allFiles: Seq[File] = matching(DependencyFilter.allPass)
 
   /** Obtains all successfully retrieved files in configurations, modules, and artifacts matching the specified filter. */
-  def matching(f: DependencyFilter): Seq[File] = select0(f).distinct
+  private[sbt] def matching(f: DependencyFilter): Seq[File] = select0(f).distinct
 
   /** Obtains all successfully retrieved files matching all provided filters.  An unspecified argument matches all files. */
-  def select(configuration: ConfigurationFilter = configurationFilter(), module: ModuleFilter = moduleFilter(), artifact: ArtifactFilter = artifactFilter()): Seq[File] =
+  def select(
+    configuration: ConfigurationFilter = configurationFilter(),
```
eed3si9n: I think we should avoid default args.

eed3si9n (changes requested):

librarymanagement/src/main/scala/sbt/librarymanagement/RichUpdateReport.scala
dwijnand: Is overloading ok? 3 versions with (undefaulted) ConfigurationFilter, ModuleFilter and ArtifactFilter?

My motivation is (attempts at) source-compatibility with 0.13.

dwijnand: +1 version with all 3 undefaulted too. That would cover the 3 use-cases I see in Scala.js (both its build and its sbt plugins).

eed3si9n: Yea. Overloading is fine.

eed3si9n (approved):

PR #80, opened 2017-04-04
[fport] Maven version range improvement

eed3si9n:
This is a forward port of https://github.com/sbt/sbt/pull/2959

Previously, when the dependency resolver (Ivy) encountered a Maven version range such as `[1.3.0,)`
it would go out to the Internet to find the latest version.
This would result to a surprising behavior where the eventual version keeps changing over time
*even when there's a version of the library that satisfies the range condition*.

This changes to some Maven version ranges would be replaced with its lower bound
so that when a satisfactory version is found in the dependency graph it will be used.
You can disable this behavior using the JVM flag `-Dsbt.modversionrange=false`.

Fixes sbt/sbt#2954
Ref sbt/sbt#2291 / sbt/sbt#2953

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 tmp/
 target/
+__pycache__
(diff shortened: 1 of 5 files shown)


eed3si9n:
Tests passed in https://travis-ci.org/sbt/librarymanagement/builds/218618890.

PR #87, opened 2017-04-25
Reformat with scalafmt

jvican:
Same as https://github.com/sbt/sbt/pull/3125.

PR #88, opened 2017-04-26
Fix sbt/sbt#1518: Handle cross-versioned exclusions

jvican:
The following PR does two things:

* Removes the unnecessary `SbtExclusionRule` that was introduced to
  exclude artifacts at the project level (and not the dependency level).
  This change was done in an independent class to avoid breaking
  bincompat in 0.13.x series.
* Handle exclusion rules correctly, propagating the cross version to the
  exclusions of the dependencies.

To fix sbt/sbt#1518, this PR takes the avenue taken in
`SbtExclusionRule`, it accepts `GroupArtifactID` which should be the
preferred way to specify dependencies from now on. Unlike
`SbtExclusionRule`, it also supports `ModuleID` for those that want to
exclude a concrete dependency.

`InclExcl` did not have any tests. The following commit also adds a
testing suite for it, showing how the issue it's fixed and how you
should use `exclude` if you're calling directly `ExclusionRule` instead
of passing in `GroupArtifactID` and `ModuleID`.


jvican:
Note that this depends on #87, that adds formatting and it's currently failing since I'm redesigning the whole formatting workflow. As soon as that's passing and merged, I'll rebase.

jvican:
This has been rebased on top of the latest formatting changes.

jvican:
This is rebased and ready for review. Mind having a look @eed3si9n @dwijnand ?

eed3si9n (approved):

mkurz:
Does this really work with sbt 1.9.7? For me it seems not...

I have a dependency:
```sbt
val atomRenderer = ("com.gu" %% "atom-renderer" % "1.2.0")
```
Now I want to exclude a dependency and it only works when appending `_2.13` to it, everything else fails:
```sbt
.exclude("com.typesafe.play", "twirl-api_2.13") // works

.exclude("com.typesafe.play", "twirl-api") // does NOT work

.excludeAll("com.typesafe.play" %% "twirl-api") // does NOT work
```

eed3si9n:
Created https://github.com/sbt/sbt/issues/7454 to discuss 1.9.x.

PR #90, opened 2017-04-26
Fix sbt/sbt#2982: Add a parallel Ivy engine

jvican:
This is a port of https://github.com/sbt/sbt/pull/2992.

Original description of the feature:

```
Ivy downloads have traditionally been single-threaded.

Parallel downloads are a must for a modern build tool. This commit
builds upon the work done by Josh Suereth in the branch
sbt/ivy-parallel-download-artifact.

To avoid adding external dependencies, it uses the Scala parallel
collections. If maintainers consider that is worth it to use a more
modern and appropriate approach, like Task, I'm happy to reimplement
the features with it.
```

Co-authored-by: Josh Suereth <joshua.suereth@gmail.com>

### DIFF
--- a/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/ParallelResolveEngine.scala
+++ b/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/ParallelResolveEngine.scala
@@ -1,0 +1,102 @@
+package sbt.internal.librarymanagement.ivyint
+
+import org.apache.ivy.core.event.EventManager
+import org.apache.ivy.core.event.download.PrepareDownloadEvent
+import org.apache.ivy.core.module.descriptor.Artifact
+import org.apache.ivy.core.report._
+import org.apache.ivy.core.resolve._
+import org.apache.ivy.core.sort.SortEngine
+import org.apache.ivy.util.Message
+import org.apache.ivy.util.filter.Filter
+
+import scala.collection.parallel.mutable.ParArray
+
+private[ivyint] case class DownloadResult(dep: IvyNode,
+                                          report: DownloadReport,
+                                          totalSizeDownloaded: Long)
+
+/** Define an ivy [[ResolveEngine]] that resolves dependencies in parallel. */
+private[sbt] class ParallelResolveEngine(settings: ResolveEngineSettings,
+                                         eventManager: EventManager,
+                                         sortEngine: SortEngine)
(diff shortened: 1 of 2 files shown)


jvican:
Thanks to this PR, the CI tests run in 90 seconds instead of 150. This is ready for review.

eed3si9n (approved):

jameskoch2:
As the author of a custom Resolver I wanted to poke around the thread-safety here, using 1.0.1 sources.

It looks to me like there's a small amount of mutable state inside most stock Resolvers' download() methods, by virtue of their reliance upon [BasicResolver.download](https://github.com/sbt/ivy/blob/2.3.0/src/java/org/apache/ivy/plugins/resolver/BasicResolver.java#L832).

I haven't seen this cause any actual issues in my testing, but thought I'd mention it. It looks like this would only impact logging of failed download attempts (based on quick read of unfamiliar code).

eed3si9n:
@jameskoch2 Thanks for the comment! I am in fact seeing some unexplainable NPEs and inconsistent failure logging, especially with fresh Ivy cache, so maybe there's a race condition.

librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/ParallelResolveEngine.scala
```diff
@@ -0,0 +1,102 @@
...
+        case DownloadStatus.SUCCESSFUL =>
+          size + artifactReport.getSize
+        case DownloadStatus.FAILED =>
+          val artifact = artifactReport.getArtifact
+          val mergedAttribute = artifact.getExtraAttribute("ivy:merged")
+          if (mergedAttribute != null) {
+            Message.warn(s"\tMissing merged artifact: $artifact, required by $mergedAttribute.")
+          } else {
+            Message.warn(s"\tDetected merged artifact: $artifactReport.")
+            resolver.reportFailure(artifactReport.getArtifact)
```
dwijnand, line 93: Shouldn't this be outside the if/else block?

dwijnand: I guess not, it's identical to the original:

https://github.com/sbt/ivy/blob/b18f59ea3bc914a297bb6f1a4f7fb0ace399e310/src/java/org/apache/ivy/core/resolve/ResolveEngine.java#L392-L399

PR #92, opened 2017-05-05
Add offline mode to `UpdateConfiguration`

jvican:
The following commit tries to address the well-known issue that sbt
cannot be used in offline mode. In order to enable that use case, this
commit adds support for a flag in update configuration called `offline`
that users can change as they wish (and that sbt will expose via
settings).

It adds tests to check that the resolution uses the caches instead of
trying to resolve from the Internet. Unfortunately, ivy does not expose
a way to know whether a resolution was made from the cache or the
Internet, so the test case invents a metric to check that resolution
indeed happens from cache.

In order to benefit from this 100%, we need to update to ivy 2.4.0 or
cherry-pick a commit because a major issue in `useCacheOnly` has been
fixed: https://issues.apache.org/jira/browse/IVY-1515.

In short, this is good for the dependency lock file too. Since we can
make sure that once we have downloaded and resolved all the dependencies
locally, we do resolve from the cache.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -19,6 +19,7 @@ def commonSettings: Seq[Setting[_]] = Seq(
   mimaPreviousArtifacts := Set(), // Some(organization.value %% moduleName.value % "1.0.0"),
   publishArtifact in Compile := true,
   publishArtifact in Test := false,
+  parallelExecution in Test := false,
   commands += Command.command("scalafmtCheck") { state =>
     sys.process.Process("git diff --name-only --exit-code").! match {
       case 0 => // ok
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -729,7 +729,6 @@
             { "name": "resolvers",            "type": "sbt.librarymanagement.Resolver*"            },
             { "name": "otherResolvers",       "type": "sbt.librarymanagement.Resolver*"            },
             { "name": "moduleConfigurations", "type": "sbt.librarymanagement.ModuleConfiguration*" },
-            { "name": "localOnly",            "type": "boolean"                                    },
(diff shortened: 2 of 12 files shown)


jvican:
This fixes: https://github.com/sbt/sbt/issues/2331 and https://github.com/sbt/sbt/issues/290.
Partially addresses https://github.com/sbt/sbt/wiki/User-Stories:--Offline-mode-and-Dependency-Locking, awaiting changes to upstream sbt.

jvican:
Review by @eed3si9n.
  
P.S. The difference with the current `offline` is that the new implementation will not try to download from the Internet if the artifacts cannot be resolved in the cache. So we can actually remove the previous `offline` implementation.

jvican:
Tests are now passing. This can be reviewed + merged. Ivy version updated too.

dwijnand:
```
[info] OfflineModeSpec:
[info] Offline update configuration
[info] - should reuse the caches when offline is enabled *** FAILED ***
[info]   offlineResolution.isRight was false (OfflineModeSpec.scala:39)
[info] - should reuse the caches when offline and cached resolution are enabled
[info] - should fail when artifacts are missing in the cache
[info] - should fail when artifacts are missing in the cache for cached resolution *** FAILED ***
[info]   failedOfflineResolution.isLeft was false (OfflineModeSpec.scala:61)
```

jvican:
This happened in a change I submitted after the comment. Thanks for the heads up @dwijnand.

jvican:
Tests were running in parallel... This option has been disabled, since this is not safe with ivy. Everything is green now and I'm done with the changes in this PR.

jvican:
ping @eed3si9n to merge / review

eed3si9n (approved):

eed3si9n:
Merged! Thanks for the contribution.

PR #94, opened 2017-05-09
Add a frozen mode to update configuration

jvican:
The frozen mode is a new addition for the dependency lock file.

Under this mode, ivy will make sure that transitive dependencies are disabled
at the resolver level, and not at the dependency level. So any attempt to
resolve transitively will fail -- the frozen mode assumes explicitly listed
dependencies.

This depends on https://github.com/sbt/librarymanagement/pull/92.
I'll rebase when the dependent is merged.


jvican:
More information on setting false to `checkIfChanged` is in the commit message. This is supposed to bring a slight performance improvement. Check ivy docs in commit message.

jvican:
I think this is not useful for the moment being.

PR #96, opened 2017-05-10
Make `SbtChainResolver` readable

jvican:
This is the most important part of the sbt ivy management and it's
almost unreadable in the previous shape...

This attempts to make a rewrite without any semantic change.

### DIFF
--- a/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
+++ b/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
@@ -255,6 +255,11 @@ private[sbt] case class SbtChainResolver(
      * This is a custom sbt chain operation that produces better error output and deals with
      * cases that the conventional ivy resolver does not. It accumulates the resolution of
      * several resolvers and returns the module which fits the provided resolution strategy.
+     *
+     * These are the differences with regard to the default ivy [[ChainResolver]]:
+     *   1. It skips resolution if "return first" is set to true.
+     *   2. It skips resolution if a previously resolved or cached resolution is found.
+     *   3. It always checks all the resolvers and compares timestamps for changing dependencies.
      */
     def getDependency(dd: DependencyDescriptor, data0: ResolveData): ResolvedModuleRevision = {
       val isDynamic = dd.isChanging || IvySbt.isChanging(dd.getDependencyRevisionId)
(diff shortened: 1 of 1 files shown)


jvican:
I'm doing this as a first step to implement the "only from this resolver" resolution.

jvican:
IMO, the best is to merge this and I will open another PR to follow up with the "only from this resolver" resolution feature.

eed3si9n:
> This is the most important part of the sbt ivy management..

I would be uncomfortable merging this without running it against sbt's dependency management related scripted tests.

jvican:
You're right. I'm checking if this works locally.
  
I'm trying hard not to rewrite core functionality since it takes time and it's not easy, but I could not avoid doing these changes... I could barely read the previous implementation. I think now it's clearer.

jvican:
I've just checked and everything is green on my side.

librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
```diff
@@ -71,188 +72,235 @@ private[sbt] case class SbtChainResolver(
...
+    /** Gets the dependency for a given descriptor with the pertinent resolve data.
+     *
+     * This is a custom sbt chain operation that produces better error output and deals with
+     * cases that the conventional ivy resolver does not. It accumulates the resolution of
+     * several resolvers and returns the module which fits the provided resolution strategy.
+     *
+     * These are the differences with regard to the default ivy [[ChainResolver]]:
+     *   1. It skips resolution if "return first" is set to true.
+     *   2. It skips resolution if a previously resolved or cached resolution is found.
+     *   3. It always checks all the resolvers and compares timestamps for changing dependencies.
```
eed3si9n: Technically, it will traverse all resolvers only if `updateOptions.latestSnapshots` is `true` (by default it's set to `true` I think).

The original comment that I wrote in 8e682691c2bc820640fd14ceea14b33779a2351b says `"regardless of what the "latest strategy" is set, and look for the published date or the module descriptor to sort them."` This is reference to Ivy's [latest-strategies](http://ant.apache.org/ivy/history/2.3.0/settings/latest-strategies.html) setting, which by default is set to **latest-revision**. It might be useful for future developer if we link to sbt/sbt#1520.

This also points to the tradeoff of correctness vs time when using `-SNAPSHOT`.

eed3si9n (changes requested):

librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
jvican: Thanks for all the history, it's been useful. I've added a new commit with this information. Let's merge this when it's green, I have too many PRs depending on this one :smile:.

eed3si9n (approved):
LGTM

jvican:
Can we merge and I rebase the rest of PRs?

jvican:
ping @eed3si9n. Sorry for the pings, but cannot move forward if these PRs are not merged.

PR #97, opened 2017-05-11
Allow to define concrete resolvers for dependencies

jvican:
This depends on https://github.com/sbt/librarymanagement/pull/96/files.
    
Sometimes, for predictability and performance, we may be interested in
specifying the concrete resolver that a `ModuleID` should use.

This patch achieves this by adding a new field to `UpdateOptions` and
then getting this information from the `SbtChainResolver`, that will
select the concrete resolver for a given dependency descriptor.

Why is this useful? Well, two reasons:

* Predictable behaviour. We have the guarantee that an artifact only
  comes from a concrete resolver.
* Resolution speedup. Around 1/3 or 1/2 times faster than normal
  resolution in a moderate test case scenario. If there is a lot of
  latency or network connection is poor, speedups will be higher.

Some results of tests I have just ran:

```
NORMAL RESOLUTION TIME 1790
FASTER RESOLUTION TIME 1054
```

```
NORMAL RESOLUTION TIME 2078
FASTER RESOLUTION TIME 1055
```

Lots of projects can benefit from this option, as well as organizations
and companies. This will eventually integrate with the dependency lock
file, but can be used independently of it.

### DIFF
--- a/librarymanagement/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
+++ b/librarymanagement/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
@@ -349,6 +349,8 @@ private[sbt] object IvySbt {
     val mainChain = makeChain("Default", "sbt-chain", resolvers)
     settings.setDefaultResolver(mainChain.getName)
   }
+
+  // TODO: Expose the changing semantics to the caller so that users can specify a regex
   private[sbt] def isChanging(dd: DependencyDescriptor): Boolean =
     dd.isChanging || isChanging(dd.getDependencyRevisionId)
   private[sbt] def isChanging(module: ModuleID): Boolean =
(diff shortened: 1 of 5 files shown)


jvican:
This has been rebased and is ready for review.

jvican:
Tests are green.

librarymanagement/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
```diff
@@ -349,6 +349,8 @@ private[sbt] object IvySbt {
     val mainChain = makeChain("Default", "sbt-chain", resolvers)
     settings.setDefaultResolver(mainChain.getName)
   }
+
+  // TODO: Expose the changing semantics to the caller so that users can specify a regex
```
jvican, line 353: I have followed up with: https://github.com/sbt/librarymanagement/issues/101.

librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
```diff
@@ -49,21 +51,26 @@ final class UpdateOptions private[sbt] (
...
+
   private[sbt] def copy(
       circularDependencyLevel: CircularDependencyLevel = this.circularDependencyLevel,
       interProjectFirst: Boolean = this.interProjectFirst,
       latestSnapshots: Boolean = this.latestSnapshots,
       consolidatedResolution: Boolean = this.consolidatedResolution,
       cachedResolution: Boolean = this.cachedResolution,
-      resolverConverter: UpdateOptions.ResolverConverter = this.resolverConverter
+      resolverConverter: UpdateOptions.ResolverConverter = this.resolverConverter,
+      moduleResolvers: Map[ModuleID, Resolver] = this.moduleResolvers
```
eed3si9n, line 64: moduleResolvers should be added to equals and hashCode.

librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
```diff
@@ -23,7 +23,9 @@ final class UpdateOptions private[sbt] (
     // If set to true, use cached resolution.
     val cachedResolution: Boolean,
     // Extension point for an alternative resolver converter.
-    val resolverConverter: UpdateOptions.ResolverConverter
+    val resolverConverter: UpdateOptions.ResolverConverter,
+    // Map the unique resolver to be checked for the module ID
+    val moduleResolvers: Map[ModuleID, Resolver]
```
eed3si9n, line 28: Should this be an immutable Map?

eed3si9n (changes requested):
Looks good overall.
Minor comments.

librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
jvican: This is a very good catch.

librarymanagement/src/main/scala/sbt/librarymanagement/UpdateOptions.scala
jvican: Yes, it should be immutable. We don't transform it internally and we don't want their contents to be exposed and modified from the outside. We just use it to represent some data that we use once inside sbt' chain resolver.

jvican:
Speedup is 3x in CI.
```
NORMAL RESOLUTION TIME 960

FASTER RESOLUTION TIME 361
```

jvican:
I just fixed the remaining bit. PR is green.

eed3si9n (approved):

PR #98, opened 2017-05-11
Fix sbt/sbt#2650: Degrade log level and clarify

jvican:
Ref sbt/sbt#2650

To ensure that SNAPSHOTs are always the latest, we go through all the
resolved modules and check their timestamps. Good.

However, if in the process of reparsing/redownloading the module
descriptor we fail (or it's not found in that resolver at all), then we
cannot refresh the resolved module that could have been internalized or
heavily cached in memory by ivy. We do this for correctness.

This patch does two things:

1. Adds more comments.
2. Warns only when there are parsing errors in ivy files.
3. Adds debug info in the rest of the cases.

This removes the pain of seeing `Unable to parse` that could be caused by
other reasons that are not related to parsing at all and which would not
affect the algorithm at hand. For instance, if we get a URLResource,
that's totally fine -- there is no way we could parse the ivy file. Ivy
uses URLResources in several cases where the artifact origin URL is
specified.

### DIFF
--- a/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
+++ b/librarymanagement/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
@@ -306,34 +306,45 @@ private[sbt] case class SbtChainResolver(
     }
   }
 
-  // Ivy seem to not want to use the module descriptor found at the latest resolver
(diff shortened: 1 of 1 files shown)


jvican:
ping @eed3si9n

eed3si9n (approved):

jvican:
Rebased and tests are green.

PR #100, opened 2017-05-11
Add a frozen mode to update configuration

jvican:
The frozen mode is used by the dependency lock file.

It makes sure that the resolution is always intransitive and that ivy
does not check for changed dependencies and stores that information in
the resolved ivy files. Following the ivy documentation, the last change
may bring a slight performance improvement:

http://ant.apache.org/ivy/history/latest-milestone/use/resolve.html

### DIFF
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -635,7 +635,8 @@
         { "name": "missingOk",      "type": "boolean"                                               },
         { "name": "logging",        "type": "sbt.librarymanagement.UpdateLogging"                   },
         { "name": "artifactFilter", "type": "sbt.librarymanagement.ArtifactTypeFilter"              },
-        { "name": "offline",        "type": "boolean"                                               }
+        { "name": "offline",        "type": "boolean"                                               },
+        { "name": "frozen",         "type": "boolean"                                               }
       ]
     },
     {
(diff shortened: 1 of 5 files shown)


jvican:
Review by @eed3si9n.

jvican:
This is passing and gives us pretty good performance results. It's more than 2x faster.
  
From the CI, we get these results:
* Normal resolution: `:: resolution report :: resolve 185ms :: artifacts dl 1961ms`.
* Frozen resolution: `:: resolution report :: resolve 76ms :: artifacts dl 1635ms`.
  
Yay.

jvican:
ping @eed3si9n

eed3si9n (changes requested):
I think the general direction is fine, but I'm not sure about the term "frozen." As seen in the debate around eviction warning, I think we should try to stick to some terms that are standard.
Would it make sense to call it "locked" mode?

jvican:
@eed3si9n Frozen resolution is the term used by Pants. I think it's the clearest one, since "locked" could mean something else. To me, frozen describes the fact that all dependencies are frozen (no transitive resolution will happen) and that artifacts are not supposed to change those transitive dependencies (that's why `checkDeps` is disabled.

eed3si9n:
The only reference I could find was the source code (https://github.com/pantsbuild/pants/blob/master/src/python/pants/backend/jvm/ivy_utils.py). Is there a page where it is explained?

jvican:
Yes, [here](https://rbcommons.com/s/twitter/r/3555/diff/6#index_header).

eed3si9n (approved):

PR #103, opened 2017-05-18
Add missing CrossVersion.disabled

cunei:


eed3si9n:
I want to clarify that this is "missing" in the sense of consistency, but not missing from migration from sbt 0.13, correct? `def full` was there because `Full(...)` used to wrap a function.

https://github.com/sbt/sbt-zero-thirteen/blob/596702f9297b364ef0f77b01758362875d975488/ivy/src/main/scala/sbt/CrossVersion.scala

cunei:
The Full/Binary/Disabled classes and objects moved with the switch to Contraband: they used to live under CrossVersion, and now are directly in sbt.librarymanagement.xxx. Since Disabled did not have an associated `disabled` method, what used to be CrossVersion.Disabled (and not CrossVersion.disabled) can now only be referred to as `sbt.librarymanagement.Disabled()`. If people used CrossVersion.Disable, that is going to break in any case when upgrading. Therefore, it makes sense to offer at least a CrossVersion.disabled, for consistency with the others (.full, .binary, etc).

eed3si9n:
ok. So should we also add `.disabled` as well to sbt 0.13.x so someone can cross build?

eed3si9n (approved):

cunei:
I think so; either that, or adding a CrossVersion.Disabled to 1.0.

eed3si9n:
Yea. Let's do that.

Could you add `def Disabled: CrossVersion = sbt.librarymanagement.Disabled()` in addition?

dwijnand:
Could you make it a val and not deprecated so it can be in pattern matching, please?

* https://github.com/sbt/sbt/commit/7fcfec8b8e2b58b33b028606ecafc8ae03f7cdd4#diff-d6dcabccfa076d9f9b74ace121e616d0L72
* https://github.com/scala-js/scala-js/compare/master...dwijnand:sbt1?expand=1#diff-44651170054d48e2ec5d9afe83219da5R38

cunei:
As things stand, pattern matching will certainly break for some people, considering that CrossVersion.Full, CrossVersion.Binary, etc do not exist any longer. A better solution would be to add aliases to all of them within CrossVersion, pointing to the new locations. Maybe using `type` ?

dwijnand:
Yeah the migration to Contraband maimed the migration.

I agree to adding both type and val aliases for Full/Binary/Disabled to avoid unnecessary differences.

librarymanagement/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -13,33 +19,40 @@ abstract class CrossVersionFunctions {
...
-
-  /** Same as .disabled, for compatibility with 0.13 */
-  val Disabled: CrossVersion = sbt.librarymanagement.Disabled()
+  val disabled: CrossVersion = LDisabled()
+
+  /** Compatibility with 0.13 */
+  val Disabled: CrossVersion = disabled
+  val Binary: CrossVersion = binary
+  val Full: CrossVersion = full
+  val Patch: CrossVersion = patch
```
dwijnand, line 40: These must be of the target's type, not CrossVersion, to function with pattern matching:

```scala
val Disabled  = LDisabled
val Binary    = LBinary
val Full      = LFull
val Patch     = LPatch
type Binary   = LBinary
type Full     = LFull
type Patch    = LPatch
```

librarymanagement/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -13,33 +19,40 @@ abstract class CrossVersionFunctions {
...
+  type Disabled = LDisabled
```
dwijnand, line 41: There is no type `sbt.CrossVersion.Disabled` in sbt 0.13. There is only the singleton type `sbt.CrossVersion.Disabled.type`.

dwijnand (changes requested):

dwijnand:
@cunei Would you like me to finish this off?

cunei:
@dwijnand Yes, please go ahead; I'll be stuck on dbuild for another while. Thanks.

dwijnand:
Superseded by #121.

PR #105, opened 2017-05-26
Implement `managedChecksums` in ivy

jvican:
Managed checksums tells ivy to forget about checking checksums for jar
files and just persist them in the cache.

The user that enables that option will take care of verifying they are
correct.

This is a big deal because:
  
1. Ivy takes *a lot of time* checking timestamps for big jars, and does it
sequentially. The user (sbt) can do better by running these checks in parallel,
speeding up the whole resolution process!
2. The fact that the sha files are not present in the cache means that build
tools cannot check if a checksum is correct without preprocessing the jar.

Note that the user is responsible for keeping the cache consistent. If the
checksum is not correct, the user should report it and overwrite the
checksum file with the valid digest.

### DIFF
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -267,21 +267,32 @@
         { "name": "organization",      "type": "String"                                                                                               },
         { "name": "name",              "type": "String"                                                                                               },
         { "name": "revision",          "type": "String"                                                                                               },
-        { "name": "configurations",    "type": "Option[String]",                      "default": "None",                             "since": "0.0.1" },
-        { "name": "isChanging",        "type": "boolean",                             "default": "false",                            "since": "0.0.1" },
-        { "name": "isTransitive",      "type": "boolean",                             "default": "true",                             "since": "0.0.1" },
-        { "name": "isForce",           "type": "boolean",                             "default": "false",                            "since": "0.0.1" },
-        { "name": "explicitArtifacts", "type": "sbt.librarymanagement.Artifact*",     "default": "Vector.empty",                     "since": "0.0.1" },
-        { "name": "inclusions",        "type": "sbt.librarymanagement.InclExclRule*", "default": "Vector.empty",                     "since": "0.0.1" },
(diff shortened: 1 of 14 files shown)


jvican:
Review by @eed3si9n.

jvican:
I forgot to mention that this also adds a field `checksum` to `ModuleID`, which is super useful for the dependency lock file and for those users that want to explicitly say what the checksum a module is (a la Pants).

jvican:
Tests are green.

librarymanagement/src/main/contraband/librarymanagement.json
```diff
@@ -267,23 +267,34 @@
...
+        { "name": "explicitArtifacts", "type": "sbt.librarymanagement.Artifact*",        "default": "Vector.empty",                     "since": "0.0.1" },
+        { "name": "inclusions",        "type": "sbt.librarymanagement.InclExclRule*",    "default": "Vector.empty",                     "since": "0.0.1" },
+        { "name": "exclusions",        "type": "sbt.librarymanagement.InclExclRule*",    "default": "Vector.empty",                     "since": "0.0.1" },
+        { "name": "extraAttributes",   "type": "Map[String, String]",                    "default": "Map.empty",                        "since": "0.0.1" },
+        { "name": "crossVersion",      "type": "sbt.librarymanagement.CrossVersion",     "default": "sbt.librarymanagement.Disabled()", "since": "0.0.1" },
+        { "name": "branchName",        "type": "Option[String]",                         "default": "None",                             "since": "0.0.1" },
+        { "name": "checksum",          "type": "Option[sbt.librarymanagement.Checksum]", "default": "None",                             "since": "0.0.1" }
```
eed3si9n: I think this is at a wrong layer. First, a module may consist of many artifacts.
Second, it's an addressing mechanism, so you won't know the checksum up front.

eed3si9n (changes requested):
As noted in the Gitter chat, a better optimization around checksum is to utilize HTTP headers when available (Maven Central provides `X-Checksum-SHA1`).

Also ModuleID is not the right place, I think.

eed3si9n (approved):

eed3si9n:
Merged! Thanks for the contribution.

jvican:
I have just opened an issue to keep track of your suggestion: https://github.com/sbt/librarymanagement/issues/108.

PR #109, opened 2017-05-29
Fix sbt/sbt#3226: Use `ConvertResolver.default`

jvican:
Sbt expects this to be the default that is used for caching update.


jvican:
We should revert this change then.

jvican:
Oh, thought this was merged. Closing, this is not necessary.

PR #112, opened 2017-06-09
Change the sbt API/binary version for sbt 1.x to be 1.0

dwijnand:


librarymanagement/src/main/scala/sbt/internal/librarymanagement/cross/CrossVersionUtil.scala
```diff
@@ -13,79 +13,81 @@ object CrossVersionUtil {
...
   def isFull(s: String): Boolean = (s == trueString) || (s == fullString)
+
   def isDisabled(s: String): Boolean =
     (s == falseString) || (s == noneString) || (s == disabledString)
+
   def isBinary(s: String): Boolean = (s == binaryString)
 
-  private lazy val intPattern = """\d{1,10}"""
-  private lazy val basicVersion = """(""" + intPattern + """)\.(""" + intPattern + """)\.(""" + intPattern + """)"""
+  private val intPattern = """\d{1,10}"""
```
eed3si9n: It might be a good opportunity to switch to using `Long` instead of `Int`.
`Int` can run out of digits when somone tries to encode date and time.
Basically all occurrence of `Int` should be `Long`.

See also https://github.com/sbt/sbt/issues/3011

eed3si9n (changes requested):

eed3si9n (approved):

PR #113, opened 2017-06-10
define def config(...) as a macro to capture the Scala identifier

eed3si9n:
In preparation to forward porting https://github.com/sbt/sbt/pull/3255, this changes the implementation of `def config(name: String)` so it captures the LHS as the `id` field in Configuration.
The constructor now enforces that the `id` (the Scala identifier for val) starts with upper case, like `Compile` and `Test`.

```scala
lazy val Compile = config("compie")
```

https://github.com/sbt/sbt/issues/2890

### DIFF
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -13,7 +13,7 @@
         { "name": "type",            "type": "String",                               "default": "Artifact.DefaultType",      "since": "0.0.1" },
         { "name": "extension",       "type": "String",                               "default": "Artifact.DefaultExtension", "since": "0.0.1" },
         { "name": "classifier",      "type": "Option[String]",                       "default": "None",                      "since": "0.0.1" },
-        { "name": "configurations",  "type": "sbt.librarymanagement.Configuration*", "default": "Vector.empty",              "since": "0.0.1" },
+        { "name": "configurations",  "type": "String*",                              "default": "Vector.empty",              "since": "0.0.1" },
         { "name": "url",             "type": "Option[java.net.URL]",                 "default": "None",                      "since": "0.0.1" },
         { "name": "extraAttributes", "type": "Map[String, String]",                  "default": "Map.empty",                 "since": "0.0.1" },
         { "name": "checksum",        "type": "Option[sbt.librarymanagement.Checksum]", "default": "None",                    "since": "0.0.1" }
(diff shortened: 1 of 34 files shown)


jvican:
Please, don't merge this. I'd like to review this PR soon and want to think a little bit about this change.

eed3si9n:
Now that we've agreed on eventual adoption of slash syntax, can I merge this soon?

jvican:
I think this is good to go, sorry for blocking. I think enforcing this in the slash syntax is key for its well-functioning.

librarymanagement/src/main/contraband/librarymanagement.json
```diff
@@ -13,7 +13,7 @@
         { "name": "type",            "type": "String",                               "default": "Artifact.DefaultType",      "since": "0.0.1" },
         { "name": "extension",       "type": "String",                               "default": "Artifact.DefaultExtension", "since": "0.0.1" },
         { "name": "classifier",      "type": "Option[String]",                       "default": "None",                      "since": "0.0.1" },
-        { "name": "configurations",  "type": "sbt.librarymanagement.Configuration*", "default": "Vector.empty",              "since": "0.0.1" },
+        { "name": "configurations",  "type": "String*",                              "default": "Vector.empty",              "since": "0.0.1" },
```
dwijnand: Please can we not make this change? This could be `ConfigKey`, or similar.

dwijnand (changes requested):

librarymanagement/src/main/contraband/librarymanagement.json
eed3si9n: Yea. I think it would be good to introduce ConfigRef in this PR.

librarymanagement/src/main/scala/sbt/librarymanagement/ConfigurationExtra.scala
```diff
@@ -80,10 +86,51 @@ abstract class ConfigurationExtra {
...
+    }
+    enclosingVal(enclosingTrees(c).toList)
+  }
+
+  def enclosingTrees(c: blackbox.Context): Seq[c.Tree] =
+    c.asInstanceOf[reflect.macros.runtime.Context]
+      .callsiteTyper
+      .context
+      .enclosingContextChain
+      .map(_.tree.asInstanceOf[c.Tree])
```
dwijnand, line 136: It's a shame to have this code copy-pasted from sbt's KeyMacro :-/

eed3si9n: Would you rather have me add util-lhs?

dwijnand: (┛◉Д◉)┛彡┻━┻

😃

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/113/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
Rebased to 1.0 and added ConfigRef.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/113/head&deploy_to_bintray=false">this link.</A>

librarymanagement/src/main/scala/sbt/librarymanagement/ConfigurationFormats.scala
```diff
@@ -0,0 +1,46 @@
+/**
+ * This code is generated using [[http://www.scala-sbt.org/contraband/ sbt-contraband]].
```
dwijnand, line 2: Why is this not in the `contraband-scala` directory?

librarymanagement/src/main/scala/sbt/librarymanagement/Configuration.scala
```diff
@@ -0,0 +1,79 @@
+package sbt
+package librarymanagement
+
+/** Represents an Ivy configuration. */
+final class Configuration private[sbt] (
```
eed3si9n, line 5: @dwijnand The ctor for Configuration is private.

eed3si9n:
Can we merge this so the review surface will be smaller for the API PR?

dwijnand (approved):

PR #117, opened 2017-06-23
Add constant CrossVersion

smarter:
sbt 1 removes CrossVersion.binaryMapped which was used in the sbt-dotty
plugin to provide a way to depend on Scala 2.x artifacts in a project
that cross-compiles between Scala 2.x and Dotty (see `withDottyCompat()` in
https://github.com/lampepfl/dotty/blob/master/sbt-dotty/src/dotty/tools/sbtplugin/DottyPlugin.scala).

Using `binaryWith` is not enough because it only allows the user to
specify a prefix and a suffix for the binary version which will always
be set to `scalaBinaryVersion`. This commit introduces a new `Constant`
kind of CrossVersion which allows the user to specify any string he
wants as a cross-version, thus making it possible to port
`withDottyCompat()` to sbt 1.

### DIFF
--- a/librarymanagement/src/main/contraband/librarymanagement.json
+++ b/librarymanagement/src/main/contraband/librarymanagement.json
@@ -151,6 +151,18 @@
           ]
         },
         {
+          "name": "Constant",
+          "namespace": "sbt.librarymanagement",
+          "target": "Scala",
+          "doc": [
+            "Cross-versions a module using the string `value`."
+          ],
+          "type": "record",
+          "fields": [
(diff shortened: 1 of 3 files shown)


dwijnand:
Hey, @smarter. If possible could you capture the use case in a test in CrossVersionTest, please? It would be mutually beneficial so we don't drop support for your use case.

smarter:
@dwijnand Sure, done!

dwijnand (approved):
Thanks. 😆

PR #118, opened 2017-06-26
In-source the modules of sbt/util required in sbt/lm

dwijnand:
The subset of sbt/util was prepared like this:

1. Move sbt.internal.util.HListFormats from util-cache to util-collection
(which isn't include here) to break the dependency between util-cache and
util-collection.

2. Move sbt.util.ShowLines out of util-collection to its own dependency to
break the dependency of util-logging on util-collection.

3. Move sbt.internal.util.Position out of util-collection to its own dependency
to break the dependency of lm on util-collection.

4. Run git filter-branch as such:

    cd /s
    git clone /d/sbt-util /d/sbt-util-4-lm
    git filter-branch --index-filter \
      'git rm --cached -qr -- . && git reset -q $GIT_COMMIT -- interface internal/util-interface util/log internal/util-logging internal/util-position internal/util-showlines cache internal/util-cache util-cache' \
      --prune-empty
    git reset --hard
    git gc --aggressive
    git prune

Then that branch was merged into the 'eat-util' branch here, the build was
configured, and the source auto-formatted.


build.sbt
```diff
@@ -55,14 +55,66 @@ lazy val lmRoot = (project in file("."))
...
+  autoScalaLibrary := false
+)
+
+// defines Java structures used across Scala versions, such as the API structures and relationships extracted by
+//   the analysis compiler phases and passed back to sbt.  The API structures are defined in a simple
+//   format from which Java sources are generated by the datatype generator Projproject
+val lmInterface = project in file("internal") / "util-interface" settings (
+  commonSettings,
+  javaOnlySettings,
+  name := "LM Interface",
```
eed3si9n: I don't know if the name LM Interface or LM Logging make sense, since it really isn't about library management.

eed3si9n (changes requested):

typesafe-tools:
To validate this pull request against other sbt modules, visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/118/head&deploy_to_bintray=false">THIS LINK.</A>

dwijnand:
See https://github.com/sbt/util/issues/58#issuecomment-312242348.

PR #123, opened 2017-07-02
Library management API

eed3si9n:
Ref https://github.com/sbt/sbt/issues/2294

The goal of Library Management API is to create an Ivy-free interface that can abstract different dependency resolution implementations such as Ivy, cached resolution, and Coursier.

Note this PR is intended to be Refactoring, in the sense that this should not introduce any change to the observable behavior.

### LibraryManagementInterface

This interface is for the dependency resolution engine authors:

```scala
/**
 * Interface for library management
 */
abstract class LibraryManagementInterface {

  /**
   * Builds a ModuleDescriptor that describes a subproject with dependencies.
   *
   * @param moduleSetting It contains the information about the module including the dependencies.
   */
  def buildModule(moduleSetting: InlineConfiguration): ModuleDescriptor

  /**
   * Resolves the given module's dependencies performing a retrieval.
   *
   * @param module The module to be resolved.
   * @param configuration The update configuration.
   * @param uwconfig The configuration to handle unresolved warnings.
   * @param log The logger.
   * @return The result, either an unresolved warning or an update report. Note that this
   *         update report will or will not be successful depending on the `missingOk` option.
   */
  def updateEither(module: ModuleDescriptor,
                   configuration: UpdateConfiguration,
                   uwconfig: UnresolvedWarningConfiguration,
                   log: Logger): Either[UnresolvedWarning, UpdateReport]

  /**
   * Publishes the given module.
   *
   * @param module The module to be resolved.
   * @param configuration The publish configuration.
   * @param log The logger.
   */
  def publish(module: ModuleDescriptor, configuration: PublishConfiguration, log: Logger): Unit

  /**
   * Makes the .ivy file for the given module.
   *
   * @param module The module to be resolved.
   * @param configuration The makeIvyFile configuration.
   * @param log The logger.
   */
  def makeIvyFile(module: ModuleDescriptor, configuration: DeliverConfiguration, log: Logger): File

  /**
   * Makes the .pom file for the given module.
   *
   * @param module The module to be resolved.
   * @param configuration The makeIvyFile configuration.
   * @param log The logger.
   */
  def makePomFile(module: ModuleDescriptor, configuration: MakePomConfiguration, log: Logger): File
}

trait ModuleDescriptor {
  def directDependenciesForWarning: Vector[ModuleID]
  def scalaModuleInfo: Option[ScalaModuleInfo]
}
```

These methods define the most general forms of performing `update` and `publish`.
If the implementation already integrates with sbt, it should be easy to implement them.

[...]

PR #124, opened 2017-07-05
Library management API

eed3si9n:
This is continuation of https://github.com/sbt/librarymanagement/pull/113 and https://github.com/sbt/librarymanagement/pull/123 on a shared branch.

Ref https://github.com/sbt/sbt/issues/2294

The goal of Library Management API is to create an Ivy-free interface that can abstract different dependency resolution implementations such as Ivy, cached resolution, and Coursier.

Note this PR is intended to be Refactoring, in the sense that this should not introduce any change to the observable behavior.

### DependencyResolution

`DependencyResolution` is the user-facing API that contains useful helper functions.
The intended audience is build users and plugin authors.

```scala
scala> import java.io.File
import java.io.File

scala> import sbt.librarymanagement._, syntax._
import sbt.librarymanagement._
import syntax._

scala> val log = sbt.util.LogExchange.logger("test")
log: sbt.internal.util.ManagedLogger = sbt.internal.util.ManagedLogger@c439b0f

scala> val lm = {
         import sbt.librarymanagement.ivy._
         val ivyConfig = InlineIvyConfiguration().withLog(log)
         IvyDependencyResolution(ivyConfig)
       }
lm: sbt.librarymanagement.DependencyResolution = sbt.librarymanagement.DependencyResolution@6a9b40f8

scala> val module = "commons-io" % "commons-io" % "2.5"
module: sbt.librarymanagement.ModuleID = commons-io:commons-io:2.5

scala> lm.retrieve(module, scalaModuleInfo = None, new File("target"), log)
res0: Either[sbt.librarymanagement.UnresolvedWarning,Vector[java.io.File]] = Right(Vector(target/jars/commons-io/commons-io/commons-io-2.5.jar, target/jars/commons-io/commons-io/commons-io-2.5.jar, target/jars/commons-io/commons-io/commons-io-2.5.jar))
```

### DependencyResolutionInterface

`DependencyResolutionInterface` is the interface dependency resolution engine authors will implement to provide the resolution feature:

```scala
/**
 * Interface for dependency resolution intended for engine authors.
 */
trait DependencyResolutionInterface {

  /**
   * Builds a ModuleDescriptor that describes a subproject with dependencies.
   *
   * @param moduleSetting It contains the information about the module including the dependencies.
   * @return A `ModuleDescriptor` describing a subproject and its dependencies.
   */
  def moduleDescriptor(moduleSetting: ModuleDescriptorConfiguration): ModuleDescriptor

  /**
   * Resolves the given module's dependencies performing a retrieval.
   *
   * @param module The module to be resolved.
   * @param configuration The update configuration.
   * @param uwconfig The configuration to handle unresolved warnings.
   * @param log The logger.
   * @return The result, either an unresolved warning or an update report. Note that this
   *         update report will or will not be successful depending on the `missingOk` option.
   */
  def update(module: ModuleDescriptor,
             configuration: UpdateConfiguration,
             uwconfig: UnresolvedWarningConfiguration,
             log: Logger): Either[UnresolvedWarning, UpdateReport]
}

trait ModuleDescriptor {
  def directDependenciesForWarning: Vector[ModuleID]
  def scalaModuleInfo: Option[ScalaModuleInfo]
}
```

These methods define the most general forms of performing `update`.
If the implementation already integrates with sbt, it should be easy to implement them.

### IvyLibraryManagement

Here's the reference implementation using Ivy.

https://github.com/eed3si9n/librarymanagement/blob/7b21353a06aab419de7892fd13cee2f4da8f26da/ivy/src/main/scala/sbt/librarymanagement/IvyLibraryManagement.scala

### Credits

This PR is co-authored by Eugene Yokota (@eed3si9n, Lightbend) and Martin Duhem (@Duhemm, Scala Center). I've squash a bunch of commits by Martin from #125 and #127.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

Duhemm:
Note: I've already started to port sbt/zinc and sbt/sbt to the new LM API described in this PR:
 - https://github.com/sbt/zinc/pull/335

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

Duhemm:
I've created a gist that shows how to resolve a given module ID with this API:
https://gist.github.com/Duhemm/0b2a6a7bd198abfda36f5412d49bdaf9

I think that there's still room for improvement. Here are some notes that I took while writing that:

 - [x] `IvyConfiguration` is in `sbt.internal`, but we need it to use this API with Ivy. Fixed in #127 
 - [x] `InlineIvyConfiguration` should also follow the builder pattern, and it requires some sane defaults. Maybe that we could also come up with a better name... See #128 
 - [ ] ~`IvyConfiguration` uses a logger, do we really need to give another one to `retrieve`? Can't it use the logger of its `LibraryManagement` instance? Does `IvyConfiguration` really need a logger?~
 - [x] `IvyPaths` is in `sbt.internal`, but we need it to instantiate `IvyConfiguration`. Fixed #128 
 - [ ] ~Both `IvyPaths` and `IvyConfiguration` need a `baseDirectory`. Couldn't they share the same? Fixed in #128~
 - [ ] We can get default resolvers using `Resolver.withDefaultResolvers`. I think that we should drop the `with` from the name. Fixed in #128 
 - [ ] ~Why do we need `resolvers` and `otherResolvers`?~
 - [ ] `Resolver.withDefaultResolvers` returns a `Seq`, but `InlineIvyConfiguration` wants a `Vector`. Fixed in #128
 - [ ] Why isn't there a way to set the configuration on an instance of `ModuleDescriptor`?
 - [ ] `def moduleDescriptor` takes a `Vector[Configuration]`, but `ModuleID` takes `Option[String]`.
 - [ ] In [this gist](https://gist.github.com/Duhemm/0b2a6a7bd198abfda36f5412d49bdaf9#file-libman-scala-L44), I need to set the configurations on the `ModuleID` to `component`. Any other configuration will fail. That's extremely confusing.

That would be great if we could get this example to become shorter.)

Duhemm:
I guess we should move everything that's specific to the Ivy implementation to the `sbt.librarymanagement.ivy` namespace. WDYT?

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
@Duhemm Here's my take on the builder pattern - https://github.com/sbt/librarymanagement/pull/124/commits/e077272df2ec1e6acf7331f867e1794a4a7628f9

eed3si9n:
@Duhemm 

> IvyConfiguration uses a logger, do we really need to give another one to retrieve? Can't it use the logger of its LibraryManagement instance? Does IvyConfiguration really need a logger?

We need logger during the setup. Task logging needs to be split for `last` to work etc.

> Both IvyPaths and IvyConfiguration need a baseDirectory. Couldn't they share the same? Fixed in #128

IvyPaths is used throughout testing, and I actually think it's a good idea to change them both at the same time, so I am keeping this one. I did move baseDirectory to external, so there are not duplicates.

> We can get default resolvers using Resolver.withDefaultResolvers. I think that we should drop the with from the name. Fixed in #128

I don't think `defaultResolvers` is good since it's returning more than the default.
`combineDefaultResolvers`? (since we can't say for sure if it's prepend or append)

> Why do we need `resolvers` and `otherResolvers`?

See Keys.scala.

>   val otherResolvers = TaskKey[Seq[Resolver]]("other-resolvers", "Resolvers not included in the main resolver chain, such as those in module configurations.", CSetting)

> Resolver.withDefaultResolvers returns a Seq, but InlineIvyConfiguration wants a Vector. Fixed in #128

We should use Vector.

> Why isn't there a way to set the configuration on an instance of ModuleDescriptor?

Not sure what you mean by this. You pass in the list of configurations to InlineConfigurations.

>  def moduleDescriptor takes a Vector[Configuration], but ModuleID takes Option[String].

I think this fine. The module descriptor has the authoritative list of configurations that it defines. ModuleID can have complex expressions like `"test->test"`.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
Rebased and squashed.

Here's demo a code:

```scala
scala> import java.io.File
import java.io.File

scala> import sbt.librarymanagement._, syntax._
import sbt.librarymanagement._
import syntax._

scala> val log = sbt.util.LogExchange.logger("test")
log: sbt.internal.util.ManagedLogger = sbt.internal.util.ManagedLogger@c439b0f

scala> val lm = {
         import sbt.librarymanagement.ivy._
         val ivyConfig = InlineIvyConfiguration().withLog(log)
         IvyLibraryManagement(ivyConfig)
       }
lm: sbt.librarymanagement.LibraryManagement = sbt.librarymanagement.ivy.IvyLibraryManagement@11c07acb

scala> val module = "commons-io" % "commons-io" % "2.5"
module: sbt.librarymanagement.ModuleID = commons-io:commons-io:2.5

scala> lm.retrieve(module, scalaModuleInfo = None, new File("target"), log)
res0: Either[sbt.librarymanagement.UnresolvedWarning,Vector[java.io.File]] =
  Right(Vector(target/jars/commons-io/commons-io/commons-io-2.5.jar,
    target/jars/commons-io/commons-io/commons-io-2.5.jar,
    target/jars/commons-io/commons-io/commons-io-2.5.jar))
```

Duhemm:
Very nice. I think that we now have a much better story in terms of user API.

If you think the change is worth it, this commit: https://github.com/Duhemm/librarymanagement/commit/460d4d766acacfcb763a89646129b349f5315863 changes the configurations in `ModuleID` to be a `Vector[String]` rather than `Option[String]`. The tests pass, but I admit that I am not 100% confident about this change as it is now. I may very well have broken lots of other things. If you think that this is an improvement, I'd be happy to work a bit more on it and making sure that I didn't break something.

I'll update my WIP PR to port Zinc to this new API, this will give us a new datapoint to assess the quality of this API.

dwijnand:
@Duhemm I like what you've done - I've always been in favour of a more typed way to define module configurations. But I think in addition to splitting the string into a seq we should type the elements, i.e not have them as strings such as "compile" and "test->test".

Also, what's the thinking behind using a `Vector("compile")` default over empty vector?

core/src/main/scala/sbt/librarymanagement/DependencyBuilders.scala
```diff
@@ -0,0 +1,67 @@
...
+}
+
+object DependencyBuilders {
+  final class Organization private[sbt] (private[sbt] val organization: String) {
+    def %(name: String) = organizationArtifact(name, Disabled())
+    def %%(name: String): OrganizationArtifactName =
+      organizationArtifact(name, CrossVersion.binary)
+
+    private def organizationArtifact(name: String, cross: CrossVersion) = {
+      nonEmpty(name, "Artifact ID")
```
dwijnand: "Name"

eed3si9n:
@Duhemm This PR should be refactoring, so the fact that the expected value on the specs are changing makes me a bit suspicious. Also `String` represents full expression such as `runtime->*;test->default(compile)` (see http://ant.apache.org/ivy/history/latest-milestone/ivyfile/dependency.html), so I think `Option[String]` is fine for now.

core/src/main/scala/sbt/librarymanagement/LibraryManagement.scala
```diff
@@ -0,0 +1,198 @@
...
+
+import java.io.File
+import sbt.util.Logger
+import sbt.io.Hash
+import sbt.librarymanagement.syntax._
+
+/**
+ * Helper mixin to provide methods for library management
+ */
+abstract class LibraryManagement extends LibraryManagementInterface {
```
dwijnand: (Forwarding my review of #123) Please could we define two different entities and use composition over inheritance?

core/src/main/scala/sbt/librarymanagement/LibraryManagementSyntax.scala
```diff
@@ -0,0 +1,24 @@
+package sbt.librarymanagement
+
+abstract class LibraryManagementSyntax extends DependencyBuilders {
```
dwijnand: Could this stay as a trait please? So it can be a mixin.

core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -6,10 +6,10 @@ package sbt.librarymanagement
 import java.io.{ IOException, File }
 import java.net.URL
 import scala.xml.XML
-import org.apache.ivy.plugins.resolver.DependencyResolver
+// import org.apache.ivy.plugins.resolver.DependencyResolver
```
dwijnand: commented code

ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
```diff
@@ -259,7 +260,7 @@ class MakePom(val log: Logger) {
         types match {
           case Nil                                     => Artifact.PomType
           case xs if xs.contains(Artifact.DefaultType) => Artifact.DefaultType
-          case x :: xs                                 => x
+          case x :: (xs @ _)                           => x
```
dwijnand, line 263: what's going on here?

core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
```diff
@@ -145,7 +145,8 @@ object VersionNumber {
         case (v1, v2)
             if (v1.size >= 3) && (v2.size >= 3) => // A normal version number MUST take the form X.Y.Z
           (v1._1.get, v1._2.get, v1._3.get, v1.tags, v2._1.get, v2._2.get, v2._3.get, v2.tags) match {
-            case (x1, y1, 0, ts1, x2, y2, 0, ts2) if ts1.nonEmpty || ts2.nonEmpty =>
+            case (x1 @ _, y1 @ _, 0, ts1, x2 @ _, y2 @ _, 0, ts2)
```
dwijnand, line 148: what with all the `@ _`?

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
```diff
@@ -142,37 +81,65 @@ object IvyActions {
...
   def publish(module: IvySbt#Module, configuration: PublishConfiguration, log: Logger): Unit = {
-    import configuration._
+    val resolverName = configuration.resolverName match {
+      case Some(x) => x
+      case _       => sys.error("Resolver name is not specified")
+    }
+
+    // Todo. Fix publish ordering https://github.com/sbt/sbt/issues/2088#issuecomment-246208872
+    val ivyFile: Option[File] =
+      if (configuration.publishMavenStyle) None
```
dwijnand, line 124: Isn't this a blocker for this PR? Does this (with the todo above) mean we no longer publish POM files?

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
```diff
@@ -441,6 +268,15 @@ object IvyActions {
       log: Logger
   )
 
+  implicit def toIvyFilter(f: ArtifactTypeFilter): IvyFilter = new IvyFilter {
+    override def accept(o: Object): Boolean = Option(o) exists {
+      case a: IArtifact => applyFilter(f, a)
+    }
+
+    def applyFilter(f: ArtifactTypeFilter, a: IArtifact): Boolean =
```
dwijnand: The f is redundant.

ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -209,7 +209,7 @@ private[sbt] class CachedResolutionResolveCache {
        else None) match {
         case Some(path) =>
           log.debug(s"parsing ${path.getAbsolutePath.toString}")
-          val ur = JsonUtil.parseUpdateReport(md, path, cachedDescriptor, log)
+          val ur = JsonUtil.parseUpdateReport( /* md, */ path, cachedDescriptor, log)
```
dwijnand: commented code

dwijnand (dismissed):

core/src/main/scala/sbt/librarymanagement/LibraryManagementSyntax.scala
eed3si9n: Is it really worth it? The trait-to-interface is still somewhat fragile (e.g. `val` must be final without return types), and I don't know who would ever need to mix this in.

ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
Duhemm: I think that's to silent the huge number of warnings. If I'm right, I would prefer `case x :: _ => ...`

eed3si9n: Avoiding the compiler warning. Could also be `_`, but I grew up writing `x :: xs`.

core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
eed3si9n: Avoiding compiler warnings.

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
eed3si9n: This is the first step towards folding `deliver` into `publish`.
I haven't changed anything with POM. I think it's handled as an artifact in Defaults.scala:

```scala
    publishArtifact in makePom := publishMavenStyle.value && publishArtifact.value,
    artifact in makePom := Artifact.pom(moduleName.value),
```

core/src/main/scala/sbt/librarymanagement/LibraryManagement.scala
eed3si9n: Yes. Sorry about that.

Duhemm:
@dwijnand I'm not super familiar with this code, but my understanding is that when there is no configuration that is explicitly set (`None`), it means `compile`. Similarly, here an empty vector means `compile`.

@eed3si9n I agree, this is just a POC. I changed the tests to compensate for the `:compile` that is always shown. Previously, we wouldn't show it if it was `None` (which implicitly means `compile`).

core/src/main/scala/sbt/librarymanagement/LibraryManagementSyntax.scala
dwijnand: We need it https://github.com/sbt/sbt/blob/1.0.x/sbt/src/main/scala/package.scala#L11

ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
dwijnand: TIL. Thank you.

core/src/main/scala/sbt/librarymanagement/LibraryManagementSyntax.scala
eed3si9n: ok. I guess I'll turn them into traits and put comments to remind future developers to use final val.

Duhemm:
`InlineConfiguration` does not follow the builder pattern. Is that on purpose?

eed3si9n:
@Duhemm 

Yes. There's a weird coupling between `module: ModuleID` and `moduleInfo: ModuleInfo`. Hopefully it's not a big deal since there are some helper functions in LibraryManagement to define one.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

core/src/main/scala/sbt/librarymanagement/LibraryManagementInterface.scala
```diff
@@ -0,0 +1,84 @@
+package sbt.librarymanagement
+
+import java.io.File
+import sbt.util.Logger
+
+/**
+ * Interface for library management intended for library management engine authors.
+ */
+abstract class LibraryManagementInterface {
```
dwijnand: Could the interfaces in this file be interfaces?

Also can we call this RetrieverInterface (or similar)?

eed3si9n: retrieve is a specific feature so a generic sounding name would be `DependencyResolutionInterface`.

dwijnand: WFM

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

core/src/main/scala/sbt/librarymanagement/LibraryManagementInterface.scala
eed3si9n: Done.

dwijnand:
LGTM, pending sbt/zinc and sbt/sbt ripple PRs.

core/src/main/scala/sbt/librarymanagement/LibraryManagementInterface.scala
```diff
@@ -0,0 +1,84 @@
...
+   * The direct dependencies of this module.
+   */
+  def directDependencies: Vector[ModuleID]
+
+  /**
+   * The information and module about the scala version that this module requires,
+   * if any.
+   */
+  def scalaModuleInfo: Option[ScalaModuleInfo]
+}
```
Duhemm, line 84: Could you please split this in 3 files (`DependencyResolutionInterface.scala`, `PublisherInterface.scala` and `ModuleDescriptor.scala`)?

Duhemm (changes requested):
See comment

core/src/main/scala/sbt/librarymanagement/LibraryManagementInterface.scala
eed3si9n: I actually thought it would be a good idea to keep all the open-ended interfaces so we can link to the file and talk about it.

Duhemm (dismissed):

Duhemm:
I forgot about `transitiveScratch`, it's still part of this API. Do we really want it there? Could we get documentation?

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
@Duhemm noticed that MakePomConfiguration was plain class, so I added https://github.com/sbt/librarymanagement/pull/124/commits/a09af23cc654a4de15eb9b53155e5a61b92f6f42. This is also rebased on config fix that exposes `Configuration.of(..)`.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/124/head&deploy_to_bintray=false">this link.</A>

eed3si9n:
Tests passing and `transitiveScratch` was ripped out, so I think this is good to go!

PR #128, opened 2017-07-11
Builder pattern for `IvyConfiguration`, remove `IvyPaths`

Duhemm:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/128/head&deploy_to_bintray=false">this link.</A>

Duhemm:
Updated example:

```scala
package lm

import java.io.File
import sbt.librarymanagement._
import sbt.librarymanagement.ivy._

object Main {

  def main(args: Array[String]): Unit = {
    val Array(org, name, version) = args
    fetch(org, name, version, Logger) match {
      case Left(_) => ???
      case Right(files) =>
        files.foreach(f => println(s" * $f"))
    }
  }

  private def getIvy(log: xsbti.Logger): IvyLibraryManagement = {
    val resolvers =
      Resolver.withDefaultResolvers(Seq.empty, true, true).toVector
    val configuration = InlineIvyConfiguration(log)
    new IvyLibraryManagement(configuration)
  }

  private def fetch(org: String,
                    name: String,
                    version: String,
                    log: xsbti.Logger) = {
    val ivy = getIvy(log)
    val target = new File("out")
    val module =
      ModuleID(org, name, version).withConfigurations(Some("component"))
    ivy.retrieve(module, None, target, log)
  }

  object Logger extends xsbti.Logger {
    def debug(msg: xsbti.F0[String]): Unit = println(s"[debug] ${msg()}")
    def error(msg: xsbti.F0[String]): Unit = println(s"[error] ${msg()}")
    def info(msg: xsbti.F0[String]): Unit = println(s"[info ] ${msg()}")
    def trace(msg: xsbti.F0[Throwable]): Unit = println(s"[trace] ${msg()}")
    def warn(msg: xsbti.F0[String]): Unit = println(s"[warn ] ${msg()}")
  }
}
```

Duhemm:
See also: https://github.com/sbt/librarymanagement/pull/128/files#diff-b6ca44df61b352f105ff9727fa60c83cR106

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/128/head&deploy_to_bintray=false">this link.</A>

core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -162,27 +162,27 @@ abstract class ResolverFunctions {
   def jcenterRepo = JCenterRepository
 
   /** Add the local and Maven Central repositories to the user repositories.  */
-  def withDefaultResolvers(userResolvers: Seq[Resolver]): Seq[Resolver] =
-    withDefaultResolvers(userResolvers, mavenCentral = true)
+  def defaultResolvers(userResolvers: Vector[Resolver]): Vector[Resolver] =
```
jvican, line 165: Why the removal of the idiomatic builder pattern name?

eed3si9n: Note that this gets mixed into the companion object Resolver. So I think @Duhemm is saying it makes it sound like it's a build pattern name but it's not, as you can see from the signature returning Vector of Resolver.

See also https://github.com/sbt/librarymanagement/pull/124#issuecomment-314407061 

> I don't think defaultResolvers is good since it's returning more than the default.
combineDefaultResolvers? (since we can't say for sure if it's prepend or append)

PR #132, opened 2017-07-20
sbt 1.0.0-RC2

eed3si9n:
Ref https://github.com/sbt/sbt/issues/3340

Blocked by lucidsoftware/neo-sbt-scalafmt#34


dwijnand:
@eed3si9n Given this is targeting the 1.0.0 where only regression bug-fixes will land I'm happy to drop scalafmt. On the 1.x branch we can still use 1.0.0-M6 with scalafmt until there's a scalafmt for 1.0.

dwijnand (approved):

dwijnand:
#133

PR #153, opened 2017-08-04
Make sure UpdateReport has a nice toString^2

dwijnand:
Fixes sbt/sbt#3357

### DIFF
--- a/core/src/test/scala/UpdateReportSpec.scala
+++ b/core/src/test/scala/UpdateReportSpec.scala
@@ -12,6 +12,7 @@ class UpdateReportSpec extends FlatSpec with Matchers {
       |	compile:
       |	org:name
       |		- 1.0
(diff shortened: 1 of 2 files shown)


core/src/test/scala/UpdateReportSpec.scala
```diff
@@ -12,6 +12,7 @@ class UpdateReportSpec extends FlatSpec with Matchers {
       |	compile:
       |	org:name
       |		- 1.0
+      |			publicationDate: Thu Jan 01 01:00:00 GMT 1970
```
eed3si9n: Let's use `calendarStringIso`, which uses ISO 8601 format like `1999-01-01T00:00:00.001Z`.

eed3si9n (dismissed):

eed3si9n:
The same argument about 1.0.0 vs 1.0.x.

PR #162, opened 2017-08-15
Include sbt plugin extra attributes in cached resolution

pvlugter:
I don't think it will be common to hit this, but something that we found in the Cinnamon build and needed to disable cached resolution. When cached resolution is enabled and a cross-built sbt plugin depends on another cross-built sbt plugin, then the resolution will reuse the cached dependency from whatever is resolved first — so you have an sbt 1.0 dependency for sbt 0.13 or the other way around.

The regular ivy cache includes the scala version and sbt version extra attributes in the path (such as `scala_2.12/sbt_1.0/` or `scala_2.10/sbt_0.13/` based on the ivy cache pattern. Add something similar for cached resolution.

Could be backported to sbt 0.13 as well.

Also, when trying this out with sbt 1.0, I needed to also set the metadata directory to enable cached resolution. Not sure if this was expected. So I needed to add:

```scala
updateOptions := updateOptions.value.withCachedResolution(true)
```

but also:

```scala
updateConfiguration := updateConfiguration.value.withMetadataDirectory(Some(path / to / dir))
```

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/CachedResolutionResolveEngine.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/CachedResolutionResolveEngine.scala
@@ -180,17 +180,25 @@ private[sbt] class CachedResolutionResolveCache {
   ): Either[ResolveException, UpdateReport] = {
     import sbt.io.syntax._
     val mrid = md.getResolvedModuleRevisionId
-    val (pathOrg, pathName, pathRevision) = md match {
+    def extraPath(id: ModuleRevisionId, key: String, pattern: String): String =
+      Option(id.getExtraAttribute(key)).fold(".")(pattern.format(_)) // "." has no affect on paths
+    def scalaVersion(id: ModuleRevisionId): String = extraPath(id, "e:scalaVersion", "scala_%s")
+    def sbtVersion(id: ModuleRevisionId): String = extraPath(id, "e:sbtVersion", "sbt_%s")
+    val (pathOrg, pathName, pathRevision, pathScalaVersion, pathSbtVersion) = md match {
       case x: ArtificialModuleDescriptor =>
         val tmrid = x.targetModuleRevisionId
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/162/head&deploy_to_bintray=false">this link.</A>

eed3si9n (approved):
Interesting. sbt 1 revealing bad assumptions.
Thanks for the fix.

ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/CachedResolutionResolveEngine.scala
```diff
@@ -180,16 +180,24 @@ private[sbt] class CachedResolutionResolveCache {
...
+         sbtVersion(tmrid))
       case _ =>
-        (mrid.getOrganisation, mrid.getName, mrid.getRevision)
+        (mrid.getOrganisation, mrid.getName, mrid.getRevision, scalaVersion(mrid), sbtVersion(mrid))
     }
     val staticGraphDirectory = miniGraphPath / "static"
     val dynamicGraphDirectory = miniGraphPath / "dynamic"
-    val staticGraphPath = staticGraphDirectory / pathOrg / pathName / pathRevision / "graphs" / "graph.json"
+    val staticGraphPath = staticGraphDirectory / pathScalaVersion / pathSbtVersion / pathOrg / pathName / pathRevision / "graphs" / "graph.json"
     val dynamicGraphPath = dynamicGraphDirectory / todayStr / logicalClock.toString / pathOrg / pathName / pathRevision / "graphs" / "graph.json"
```
pvlugter: Was meant to be added to this path too, will fix.

typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt%20validator/parambuild/?librarymanagement=pull/162/head&deploy_to_bintray=false">this link.</A>

PR #173, opened 2017-10-13
On JDK 9 call java.net.Authenticator.getDefault

### LINKED ISSUE #169: JDK9 Warning: Illegal reflective access
Hello.

I've installed JDK9 (MacOS) and using sbt 1.0.2 (via sbt-extras) I see the following when I start sbt:

```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by sbt.internal.librarymanagement.ivyint.ErrorMessageAuthenticator$ (file:/Users/richard/.sbt/boot/scala-2.12.3/org.scala-sbt/sbt/1.0.2/librarymanagement-ivy_2.12-1.0.2.jar) to field java.net.Authenticator.theAuthenticator
WARNING: Please consider reporting this to the maintainers of sbt.internal.librarymanagement.ivyint.ErrorMessageAuthenticator$
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

It doesn't cause me any problems but as requested in the output I'm "reporting this to the maintainers" (assuming I've reached the right place!).

Thank you.

dwijnand:
JDK 9 complains when java.lang.reflect.Field#setAccessible is called.
So on JDK 9 to get the default Authenticator on JDK you can just call
Authenticator.getDefault(). However lm targets JDK 8..

So on JDK 8 we do what we've always done, and on JDK 9 we use java
reflection to call Authenticator.getDefault(), which is public and
doesn't require setAccessible(true).

Fixes #169

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/ErrorMessageAuthenticator.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/ErrorMessageAuthenticator.scala
@@ -1,6 +1,7 @@
 package sbt.internal.librarymanagement
 package ivyint
 
+import java.lang.reflect.InvocationTargetException
 import java.net.{ Authenticator, PasswordAuthentication }
 
 import org.apache.ivy.util.Message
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?librarymanagement=pull/173/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n (approved):
Interesting. JDK 9 added `getDefault`.
LGTM

PR #203, opened 2018-01-14
Add a warning for resolver access denied errors.

### LINKED ISSUE #159: Access Denied error could have better error message
Hi I get this error when we dont have the password configured for my companies nexus server. Then sbt keeps retrying. Adding the password resolves the problem.
```
[error] Server access Error: Too many follow-up requests: 21 url=https://nexus.mycompany.com/nexus/content/groups/public/org/slf4j/slf4j-parent/1.7.23/slf4j-parent-1.7.23.jar
```
If I curl the server It defiantly returns a `HTTP/1.1 401 Unauthorized`.

I think retrying makes sense for other http errors but not the 401. In this case it would be better if the error clearly said `access denied` and maybe prompted the user to configure a password. This would help other users debug their problems.

I think the error comes from here, and we could add another case. https://github.com/sbt/librarymanagement/blob/1.x/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/GigahorseUrlHandler.scala#L65

PS. I'd be interested in contributing to this, but was unable to import project into intellij. I get `unresolved dependency: com.thoughtworks.paranamer#paranamer;2.8`

stephennancekivell:
To  improve usability. Resolves #159

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/GigahorseUrlHandler.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/GigahorseUrlHandler.scala
@@ -242,6 +242,8 @@ object GigahorseUrlHandler {
         Message.debug("HTTP response status: " + status + " url=" + url)
         if (status == 407 /* PROXY_AUTHENTICATION_REQUIRED */ ) {
           Message.warn("Your proxy requires authentication.")
+        } else if (status == 401) {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?librarymanagement=pull/203/head&deploy_to_bintray=false&run_sbt_community=false">this link.</A>

eed3si9n (approved):
Thanks!

PR #213, opened 2018-02-23
Cache the hashCode of Configuration

retronym:
I noticed this was showing up in profiles when SBT's task engine
was using Keys, etc (that contain Configurations) in HashMap's.

Let's cache it instead. I don't think there is a need to use a lazy
val for this, we can compute it eagerly.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/Configuration.scala
+++ b/core/src/main/scala/sbt/librarymanagement/Configuration.scala
@@ -31,7 +31,7 @@ final class Configuration private[sbt] (
         (this.transitive == x.transitive)
     case _ => false
   }
-  override def hashCode: Int = {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=1.1.x&librarymanagement=pull/213/head&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

dwijnand (approved):

eed3si9n (approved):
LGTM
Thanks!

PR #216, opened 2018-03-09
Replace abstract protected[this] copy with abstract withX methods

dwijnand:
It isn't required for `*Extra` classes to define abstract `copy` methods: they can define and use abstract `with*` methods instead.

This removes the default argument bincompat complexity of copy methods, allowing Contraband's copy methods to be declared `private[this]` again..


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.1.x&zinc=1.1.x&io=1.1.x&librarymanagement=pull/216/head&util=1.1.x&website=1.1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):

PR #218, opened 2018-03-11
Add APIs to compare sbt.librarymanagement.VersionNumbers.

tanishiking:
Addresses https://github.com/sbt/sbt/issues/3710.

This PR implements APIs to compare `sbt.librarymanagement.VersionNumber`s.

## Approach
This implementation basically follows https://semver.org/#spec-item-11.

- `Precedence is determined by the first difference when comparing each of these identifiers from left to right as follows`
- `When major, minor, and patch are equal, a pre-release version has lower precedence than a normal version`
- In pre-release versions comparing
  - `identifiers consisting of only digits are compared numerically`
  - `identifiers with letters or hyphens are compared lexically in ASCII sort order`
    - In this implementation, we compare identifiers case-insensitively because it seems more convinient.
  - `Numeric identifiers always have lower precedence than non-numeric identifiers. `
  - `A larger set of pre-release fields has a higher precedence than a smaller set, if all of the preceding identifiers are equal.`

## Questions
- However in semantic-versioning we can separate identifiers in pre-release version by hyphens or **dots**, this `VersionNumber#(un)apply` seems not support separation by dots. Is this deliberately? Or should we implement it? (Or am I misunderstanding semantic-versioning?)
- In comparing identifiers in pre-releases, we compare identifiers case-insensitively for convenience. Is this acceptable? Should we compare them case-sensitively?

## References
- https://github.com/scalacenter/scaladex/blob/6928c7076a77b50319f3413db73e6344f595455b/model/src/main/scala/ch.epfl.scala.index.model/SemanticVersion.scala
- https://github.com/sbt/contraband/blob/458e8d27657093c4c5efae8ac4e1e43ca67e1000/library/src/main/scala/sbt/contraband/VersionNumber.scala

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
+++ b/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
@@ -1,5 +1,7 @@
 package sbt.librarymanagement
 
+import scala.annotation.tailrec
+
 final class VersionNumber private[sbt] (
     val numbers: Seq[Long],
     val tags: Seq[String],
@@ -12,6 +14,11 @@ final class VersionNumber private[sbt] (
   def _4: Option[Long] = get(3)
   def get(idx: Int): Option[Long] = numbers lift idx
   def size: Int = numbers.size
+
+  def <(o: VersionNumber): Boolean = implicitly[Ordering[VersionNumber]].lt(this, o)
+  def <=(o: VersionNumber): Boolean = implicitly[Ordering[VersionNumber]].lteq(this, o)
+  def >(o: VersionNumber): Boolean = implicitly[Ordering[VersionNumber]].gt(this, o)
+  def >=(o: VersionNumber): Boolean = implicitly[Ordering[VersionNumber]].gteq(this, o)
 
   /** The vector of version numbers from more to less specific from this version number. */
   lazy val cascadingVersions: Vector[VersionNumber] =
@@ -75,6 +82,62 @@ object VersionNumber {
       case NonSpaceString(s) => Some((Vector.empty, Vector.empty, Vector(s)))
       case _                 => None
     }
+  }
+
+  implicit val versionNumberOrdering: Ordering[VersionNumber] =
+    new Ordering[VersionNumber] {
+      def compare(x: VersionNumber, y: VersionNumber): Int = {
+        compareSeq(x.numbers, y.numbers) match {
+          case 0 =>
+            comparePreRelease(x, y) match {
+              case 0      => compareSeq(x.extras, y.extras)
+              case resPre => resPre
+            }
+          case resNormal => resNormal
+        }
+      }
+    }
+
+  @tailrec
+  private def compareSeq[A: Ordering: Zero](a1: Seq[A], a2: Seq[A]): Int =
+    if (a1.isEmpty && a2.isEmpty) 0
+    else if (a1.nonEmpty && a2.isEmpty)
+      implicitly[Ordering[A]].compare(a1.head, implicitly[Zero[A]].zero)
+    else if (a1.isEmpty && a2.nonEmpty)
+      implicitly[Ordering[A]].compare(implicitly[Zero[A]].zero, a2.head)
+    else {
+      val a1head = a1.head
+      val a2head = a2.head
+      if (a1head == a2head) compareSeq(a1.tail, a2.tail)
+      else implicitly[Ordering[A]].compare(a1head, a2head)
+    }
+
+  private def comparePreRelease(v1: VersionNumber, v2: VersionNumber): Int = {
+    implicit val PreReleaseStringOrdering: Ordering[String] = new Ordering[String] {
+      def compare(x: String, y: String): Int = {
+        (x.isEmpty, y.isEmpty) match {
+          case (true, true)  => 0
+          case (false, true) => 1
+          case (true, false) => -1
+          case (false, false) =>
+            (x.matches("""\d+"""), y.matches("""\d+""")) match {
+              // Identifiers consisting of only digits are compared numerically.
+              // Numeric identifiers always have lower precedence than non-numeric identifiers.
+              // Identifiers with letters are compared case insensitive lexical order.
+              case (true, true)   => implicitly[Ordering[Int]].compare(x.toInt, y.toInt)
+              case (false, true)  => 1
+              case (true, false)  => -1
+              case (false, false) => x.toLowerCase.compareTo(y.toLowerCase)
+            }
+        }
+      }
+    }
+    val p1 = v1.tags
+    val p2 = v2.tags
+    if (p1.isEmpty && p2.isEmpty) 0
+    else if (p1.nonEmpty && p2.isEmpty) -1
+    else if (p1.isEmpty && p2.nonEmpty) 1
+    else compareSeq(p1, p2)
   }
 
   /** Strict. Checks everything. */
@@ -177,6 +240,17 @@ object VersionNumber {
         case _                                                    => false
       }
     }
+  }
+
+  private trait Zero[A] {
+    def zero: A
+  }
+  private object Zero {
+    def apply[A](z: => A): Zero[A] = new Zero[A] {
+      def zero: A = z
+    }
+    implicit val StringZero: Zero[String] = Zero[String]("")
+    implicit val LongZero: Zero[Long] = Zero[Long](0L)
   }
 }
 
--- a/core/src/test/scala/sbt/librarymanagement/VersionNumberSpec.scala
+++ b/core/src/test/scala/sbt/librarymanagement/VersionNumberSpec.scala
@@ -111,6 +111,112 @@ class VersionNumberSpec extends FreeSpec with Matchers with Inside {
     assertParsesToError(v)
   }
 
+  "VersionNumber" - {
+    "should have an ordering" in {
+      val versions = List(
+        "2.0.0",
+        "1.1.0",
+        "1.0.2",
+        "1.0.1",
+        "1.0.0"
+      )
+      val expected = List(
+        "1.0.0",
+        "1.0.1",
+        "1.0.2",
+        "1.1.0",
+        "2.0.0"
+      )
+      assertSortTo(versions, expected)
+    }
+
+    "should put a lower precedence to a pre-release version than a normal version" in {
+      val versions = List(
+        "1.0.0",
+        "1.0.1",
+        "1.0.0-SNAPSHOT",
+        "1.0.1-SNAPSHOT"
+      )
+      val expected = List(
+        "1.0.0-SNAPSHOT",
+        "1.0.0",
+        "1.0.1-SNAPSHOT",
+        "1.0.1"
+      )
+      assertSortTo(versions, expected)
+    }
+
+    "should put a higher precedence to a version with larger set of pre-release version's fields " +
+      "than a that with smaller set of the fields, if all of the preceding identifiers are equal" in {
+      val versions = List(
+        "1.0.0-alpha-beta-gamma",
+        "1.0.0-alpha-beta",
+        "1.0.0-alpha-bar",
+        "1.0.0-alpha"
+      )
+      val expected = List(
+        "1.0.0-alpha",
+        "1.0.0-alpha-bar",
+        "1.0.0-alpha-beta",
+        "1.0.0-alpha-beta-gamma"
+      )
+      assertSortTo(versions, expected)
+    }
+
+    "should put a lower precedence to a pre-release version with numerical identifiers " +
+      "than a that with non-numerical identifiers" in {
+      val versions = List(
+        "1.0.0-alpha-3",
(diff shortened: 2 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=1.x&io=1.x&librarymanagement=pull/218/head&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
```diff
@@ -14,6 +16,15 @@ final class VersionNumber private[sbt] (
...
   def size: Int = numbers.size
 
+  def <(o: VersionNumber)(implicit ord: Ordering[VersionNumber]): Boolean =
+    ord.compare(this, o) < 0
+  def <=(o: VersionNumber)(implicit ord: Ordering[VersionNumber]): Boolean =
+    ord.compare(this, o) <= 0
+  def >(o: VersionNumber)(implicit ord: Ordering[VersionNumber]): Boolean =
+    ord.compare(this, o) > 0
+  def >=(o: VersionNumber)(implicit ord: Ordering[VersionNumber]): Boolean =
+    ord.compare(this, o) >= 0
```
dwijnand: perhaps we can reuse `Ordering`'s inner `Ops` class here?

dwijnand (commented):
hi @tlockney!

thank you for your PR! I've been spending some time on it today. I started reviewing it (1 initial comment) but because I didn't personally know this part of the codebase I spent time getting to know it and cleaning it up, leading to my creating https://github.com/sbt/librarymanagement/pull/221 and https://github.com/sbt/librarymanagement/pull/222.

having done that I will return to review the work you've done here. I'm happy to rebase your work onto my refactoring if you'd like, just let me know.

tlockney:
Hey @dwijnand, I think you meant to tag @tanishiking on this. :-)

tanishiking:
Thank you for your reply @dwijnand.
Sure, your refactoring (#221 and #222) looks great to me and I'm happy to rebase my work onto them after they are done!

dwijnand:
cool, thanks. I'll ping you when (if) they're accepted.

dwijnand:
> However in semantic-versioning we can separate identifiers in pre-release version by hyphens or dots, this VersionNumber#(un)apply seems not support separation by dots. Is this deliberately? Or should we implement it? (Or am I misunderstanding semantic-versioning?)

first off, my interpretation is that in SemVer pre-release identifiers cannot be hyphen-separated, only dot-separated.

looking at `VersionNumber` it's quite problematic because `VersionNumber` doesn't, for instance, allow numbers or dots its `tags` fields.. it becomes part of the `extras` strings..

I'm not sure what we should do to fix this, but I would highlight that pre-release version ordering is pretty niche, not mentioned in sbt/sbt#3710, and in my opinion never going to work anyways because from my experience people (sbt included!) don't dot-separate their prerelease identifiers, e.g sbt version `1.0.0-RC4` rather than `1.0.0-RC.4`..

> In comparing identifiers in pre-releases, we compare identifiers case-insensitively for convenience. Is this acceptable? Should we compare them case-sensitively?

hehehe I looked the history here and found a comment from me in 2014 :)

https://github.com/semver/semver/issues/176

I agree with you: we should compare case-insensitively.

I wonder though whether we should change the implementation of `VersionNumber` equals (and therefore hashCode) to also ignore case differences in strings.

---

going to review your implementation next

core/src/main/scala/sbt/librarymanagement/Zero.scala
```diff
@@ -0,0 +1,13 @@
...
+trait Zero[A] {
+  def zero: A
+}
+
+object Zero {
+  def apply[A](z: => A): Zero[A] = new Zero[A] {
+    def zero: A = z
+  }
+  implicit val StringZero: Zero[String] = Zero[String]("")
+  implicit val LongZero: Zero[Long] = Zero[Long](0L)
```
dwijnand: I don't think we should introduce such a typeclass universally in sbt.librarymanagement.

Let's move it into VersionNumber, so it's a VersionNumber specific utility only.

dwijnand (approved):
I have one more request.

other than that the implementation looks ok.

after we can rebase I'd like to just review the test cases present and missing, and I'm sure we can merge this speedily.

dwijnand (changes requested):
misclicked approve :)

tanishiking:
Thank you for reviewing!

> first off, my interpretation is that in SemVer pre-release identifiers cannot be hyphen-separated, only dot-separated.

Oh, sorry I misunderstood SemVer. After reading https://semver.org/#spec-item-11 again, indeed it is described.

```
Precedence for two pre-release versions with the same major, minor, and patch version MUST be determined by comparing each dot separated identifier from left to right until a difference is found as follows:
```

> in my opinion never going to work anyways because from my experience people (sbt included!) don't dot-separate their prerelease identifiers, e.g sbt version 1.0.0-RC4 rather than 1.0.0-RC.4..

I agree with you! OK, in this PR, I'll leave it :)

---

> hehehe I looked the history here and found a comment from me in 2014 :)
> semver/semver#176
> I agree with you: we should compare case-insensitively.

Awesome, thanks!

Hmm... looking at discussion in https://github.com/semver/semver/issues/176 I think we should  change the implementation of `VersionNumber` equals (and therefore hashCode) to also ignore case differences in strings. (If we compare `VersionNumber` ignoring case differences).

dwijnand:
> Hmm... looking at discussion in semver/semver#176 I think we should change the implementation of `VersionNumber` equals (and therefore hashCode) to also ignore case differences in strings. (If we compare `VersionNumber` ignoring case differences).

looking again this PR is introducing an `Ordering` for `VersionNumber` based on the SemVer specification, in implicit scope. therefore I think we have to either (a) make equals (and hashCode) ignore case differences, or (b) make ordering case insensitive.

I'm not 100% sure because I don't know what it means to have methods based on `Ordering` on `VersionNumber` that are based on an implicit scope `Ordering`, with regards to its `equals`. does it need to be consistent? I think probably yes if it's so automatic (i.e no need for an import). need to do more research.

and if it needs to be consistent we should enforce that it is (with testing).

core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
tanishiking: done 07e9a2d

core/src/main/scala/sbt/librarymanagement/Zero.scala
tanishiking: done b67a16e

tanishiking:
I changed the `VersionNumber`'s `equals` method to ignore case difference 3182e4d. I'll squash these commits after rebasing this PR onto refactors :)

---

> I'm not 100% sure because I don't know what it means to have methods based on Ordering on VersionNumber that are based on an implicit scope Ordering, with regards to its equals

Sorry I couldn't catch your feedback. Which do you mean by "consistent"?

- `VersionNumber#equals`'s behavior should be "consistent" even if any instance of `Ordering[VersionNumber]` is in implicit scope.
  - If so, yes `Ordering[VersionNumber]` has no effects to `VersionNumber`'s `equal` method.
- `VersionNumber#equals`'s behavior should be "consistent" with definition of `Ordering[VersionNumber]`
  - In other words, `VersionNumber` should have property like `(v1 == v2) == (v1 <= v2 && v1 >= v2) for all VersionNumber`.
  - To do so, we need to ignore `extras` in `equals` in addition to ignore case difference, because in SemVer we ignore extras for comparing their versions.
- Or other?

core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
```diff
@@ -14,6 +16,11 @@ final class VersionNumber private[sbt] (
     else Some(numbers(idx))
   def size: Int = numbers.size
 
+  def <(o: VersionNumber): Boolean = this < o
+  def <=(o: VersionNumber): Boolean = this <= o
+  def >(o: VersionNumber): Boolean = this > o
+  def >=(o: VersionNumber): Boolean = this >= o
```
dwijnand: these definitions are circular

dwijnand:
`Ordering` extends `Equiv`, so it has an `equiv` method. if we're introducing an implicit scope `Ordering[NumberVersion]` then `NumberVersion`'s definition of `equals` should match that ordering's definition of `equiv`.

I think we should just delegate `equals` to `implicitly[Ordering[NumberVersion]].equiv(this, that)`.

tanishiking:
Thank you for the explanation. I understood!

> I think we should just delegate equals to implicitly[Ordering[NumberVersion]].equiv(this, that).

I agree with you (and tried to implement it e745970)  but to do that we have to either

- ignore `extra` field in `Ordering[VersionNumber]#compare` and `equals`
  - this follows semantic-versioning's rule.
    - > Build metadata SHOULD be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. https://semver.org/#spec-item-10
  - but I think it is not unintuitive to ignore `extra` in `equals` and this change will break the backword compatiblity for `VersionNumber#equals`.
  - if implement this way, I think we should remain the `VersionNumber.Strict#isCompatible` to compare `extra` field.
- do not ignore `extra` field in `Ordering[VersionNumber]#compare` (and `equals`).
  - like e745970
  - this breaks the semantic-versioning's rule.
  - but I think it is pretty niche to compare `extra` field.

I prefer the latter one. what do you think about this?

dwijnand:
> what do you think about this?

I'm torn.

to add to your list of options there is option 3: don't put the SemVer-honoring in implicit scope, but make it opt-in.

we should evaluate in terms of the effect it would be have on the use case presented in #3710

meanwhile my refactoring in https://github.com/sbt/librarymanagement/pull/222 is rebased and green. but @eed3si9n is currently on holiday until next week.

dwijnand:
@tanishiking please note that https://github.com/sbt/librarymanagement/pull/222 is now merged.

tanishiking:
Thank you for ping (and sorry for my late reply), rebased.

But, after considering the use case presented in sbt/sbt#3710 some days, I came to realize that we should implement an api like [semantic version selectors](https://draftin.com/documents/375100?token=rR30GmJJzi4l3BRlD-cHs8lcAcdDAXH4oTzqOWeL0CT0BNv3PZEx0g8pBkI13sQgYXTBqShZ0Ucsqek3Fn3d-aU) which is given by some package managers like [npm](https://www.npmjs.com/package/semver)  (e.g. `VersionNumber("2.12.2").satisfies(">=2.12")`), instead of this (semver-honoring) ordering api.

The reason why implementing VersionNumber's ordering wouldn't be adequate to resolve sbt/sbt#3710 is below.

- I also think `VersionNumber`'s ordering should be consistent with its equality(VersionNumber should satisfy symmetry rule)(`∀a,b ∈ VersionNumber, (a <= b && b <= a) <=> a == b`).
  - Considering the use case given in sbt/sbt#3710, i think,
    - `v >= VersionNumber("2.12")` expects `v = "x.y.z" | (x > 2) or (x == 2 and y >= 12)`
    - `v > VersionNumber("2.12")` expects `v = "x.y.z" | (x > 2) or (x == 2 and y > 12)`
    - `v <= VersionNumber("2.12")` expects `v = "x.y.z" | (x < 2) or (x == 2 and y <= 12)`
    - `v < VersionNumber("2.12")` expects `v = "x.y.z" | (x < 2) or (x == 2 and y < 12)`
      - I don't know how VersionNumber should behave if `v`'s minor and patch versions are omitted.
  - SemVer doesn't seem to specify how should we handle the version that doesn't have minor or patch versions like `2.12` or `2`. (refer https://github.com/semver/semver/issues/237 ), and this PR's semver-honoring ordering api doesn't satisfy above expectation.
  - We can change (or make it opt-in) the behavior of the VersionNumber's ordering so that it satisfies the expectation above (ignoring semver specification). But we can't avoid breaking the VersionNumber's symmetry rule if VersionNumber satisfies the expectation.
    - in addition, I think it is confusing to implement a ordering which isn't commonly specified (unlike semver).

On the other hand, semantic version selectors realize the use case presented in #3710 without introducing a ordering that is not commonly specified or would breaks symmetry rule.

dwijnand:
@tanishiking I'm sorry for leaving you hanging. In retrospect I realise that I should have forewarned you that I (and Eugene) were going to be busy and afk all of last week.

I think you're right and we should go for something like those selectors to avoid conflicting concerns.

Please let me know if you need a hand.

tanishiking:
@dwijnand It's no problem :)

> I think you're right and we should go for something like those selectors to avoid conflicting concerns.
> Please let me know if you need a hand.

OK, I'll try to implement something like that selector and close this PR, thank you for discussing about this!

PR #221, opened 2018-03-15
Cleanup VersionNumberSpec

dwijnand:


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=1.x&io=1.x&librarymanagement=pull/221/head&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

core/src/test/scala/sbt/librarymanagement/VersionNumberSpec.scala
```diff
@@ -1,150 +1,185 @@
...
-  def beParsedAs(s: String, ns: Seq[Long], ts: Seq[String], es: Seq[String]) =
-    s match {
-      case VersionNumber(ns1, ts1, es1) if (ns1 == ns && ts1 == ts && es1 == es) =>
-        (VersionNumber(ns, ts, es).toString shouldBe s)
-        (VersionNumber(ns, ts, es) shouldBe VersionNumber(ns, ts, es))
+// This is a specification to check VersionNumber and VersionNumberCompatibility.
+class VersionNumberSpec extends UnitSpec with Inside {
+  import VersionNumber.{ SemVer, SecondSegment }
+
+  version("1") { implicit v =>
```
eed3si9n: The overall looks is cleaner, but I am not really sure if I like `implicit` used just so it doesn't pass in one parameter.

dwijnand: I can change that implementation detail

eed3si9n: Another sort of meta question is: Do we make custom DSL for each spec / unit tests we write that reads like English statements? I'd much rather prefer we used something like JUnit or Minitest, if we are moving away from plain UnitSpec.

dwijnand: I'm happy to discuss and/or experiment with alternatives to UnitSpec.

but note that in this PR:
but I'm not moving away from UnitSpec in this PR, I'm just removing extreme repetition

dwijnand: I'm happy to discuss and/or experiment with alternatives to UnitSpec.

but note that I'm not moving away from UnitSpec in this PR.

I am in favour of creating helper methods to remove lots of repetition and improve legibility (and maintainability), but the fact that it reads like English statements is not my objective.

I'm not sure where the line is drawn between creating a few helper methods and removing repetition becomes a "custom DSL".

eed3si9n:
```scala
"1" should "be parsed" in beParsedAs("1", Seq(1), Seq(), Seq())
it should "breakdown" in breakDownTo("1", Some(1))
```

In the above, `beParsedAs` is an English style helper function idiomatic to ScalaTest's `FlatSpec`.

> but note that I'm not moving away from UnitSpec in this PR.

I'd say you've made our own SomethingSpec that internally uses `FlatSpec`. I really like it, but I also want to stick to an existing test framework, esp if we want others to follow it. Perhaps the style we should migrate is `FunSuite`:

```scala
  test("1") {
    val v = "1"
    assertParsesTo(v, Seq(1), Seq(), Seq())
    assertBreaksDownTo(v, Some(1))
    assertCascadesTo(v, Seq("1"))
  }
```

Less noise. WDYT?

dwijnand:
@eed3si9n I went with `FreeSpec` because `FunSuite` doesn't support nesting.

wdyt? I think the "assert" prefix is just noise, but I can live with it.

eed3si9n (approved):
LGTM once it passes Travis

PR #242, opened 2018-06-01
Remove requirement for capitalized ids.

### LINKED ISSUE #134: The `config()` macro should validate the ID it captures from the enclosing `val`
The `config()` macro captures the name of the enclosing `val` to determine the "id" of the configuration. In this snippet:
```scala
val fooBar = config("foo-bar")
```
it will capture the id `fooBar`. The macro does not complain.

Instead, *at run-time* (I mean, build load time), I get the following obscure error:
```
...
[debug]   Load.loadUnit(file:/localhome/doeraene/projects/scalajs/sbt-plugin-test/referencedCrossProject/, ...) took 5034.22581ms

[debug] Load.apply: load took 19346.912803ms

[debug] Load.apply: resolveProjects took 0.660934ms

[debug] Load.apply: finalTransforms took 140.090565ms

[debug] Load.apply: config.delegates took 1.482166ms

[error] java.lang.IllegalArgumentException: requirement failed: id must be capitalized: scalaJSProjectBaseSettings

[error]         at scala.Predef$.require(Predef.scala:277)

[error] java.lang.IllegalArgumentException: requirement failed: id must be capitalized: scalaJSProjectBaseSettings

[error] Use 'last' for the full log.

[debug] > Exec(load-failed, None, None)

[debug] > Exec(last, None, None)

Project loading failed: (r)etry, (q)uit, (l)ast, or (i)gnore? q
```
Note that this is the output *with* `last`. In my case `scalaJSProjectBaseSettings` is the `fooBar`, and it was very obscure because the `config()` call was deep within the rhs of `scalaJSProjectBaseSettings` (which was a `Seq[Setting[_]]`.

The `config()` macro should really validate the id at compile-time. There is no point in delaying this error until run-time, when there is not even a stack trace to figure out what's wrong.

frgomes:
Fix #134 :: Relax requirement for capitalized ids.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/Configuration.scala
+++ b/core/src/main/scala/sbt/librarymanagement/Configuration.scala
@@ -16,7 +16,6 @@ final class Configuration private[sbt] (
   require(name.nonEmpty, "name cannot be empty")
   require(id != null, "id cannot be null")
   require(id.nonEmpty, "id cannot be empty")
-  require(id.head.isUpper, s"id must be capitalized: $id")
 
   override def equals(o: Any): Boolean = o match {
     case x: Configuration =>


dwijnand:
No, the requirement is intended (but untested).

See https://github.com/sbt/librarymanagement/issues/134#issuecomment-393900597

PR #254, opened 2018-07-13
Cleanup after tests to avoid OOME: Metaspace in fast-run testOnly sequence

retronym:
After these changes, this project no longer exhibits the problems
discussed in sbt/sbt#4166.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -30,11 +30,7 @@ def commonSettings: Seq[Setting[_]] = Def.settings(
   inCompileAndTest(scalacOptions in console --= Vector("-Ywarn-unused-import", "-Ywarn-unused", "-Xlint")),
   publishArtifact in Compile := true,
   publishArtifact in Test := false,
-  parallelExecution in Test := false,
-  testOptions in Test += {
-    val log = streams.value.log
-    Tests.Cleanup { loader => cleanupTests(loader, log) }
-  }
+  parallelExecution in Test := false
 )
 
 val mimaSettings = Def settings (
@@ -106,6 +102,10 @@ lazy val lmCore = (project in file("core"))
     managedSourceDirectories in Compile +=
       baseDirectory.value / "src" / "main" / "contraband-scala",
     sourceManaged in (Compile, generateContrabands) := baseDirectory.value / "src" / "main" / "contraband-scala",
+    testOptions in Test += {
+      val log = streams.value.log
+      Tests.Cleanup { loader => cleanupTests(loader, log) }
+    },
     contrabandFormatsForType in generateContrabands in Compile := DatatypeConfig.getFormats,
     // WORKAROUND sbt/sbt#2205 include managed sources in packageSrc
     mappings in (Compile, packageSrc) ++= {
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.x&zinc=1.x&io=1.x&librarymanagement=pull/254/head&util=1.x&website=1.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
I am guessing you mean https://github.com/sbt/sbt/issues/4166.
I was using LM test as an example I had handy, but the OP likely had another build exhibiting the issue, so I think we should fix them either in ScalaTest or sbt.

retronym:
Yeah, this is really just a demo of how to do the cleanup in the current
version of SBT. The log4j shutdown part is arguably project specific,
though I guess SBT could do that in the users behalf by default

build.sbt
```diff
@@ -275,3 +279,22 @@ inThisBuild(Seq(
...
 def inCompileAndTest(ss: SettingsDefinition*): Seq[Setting[_]] =
   Seq(Compile, Test) flatMap (inConfig(_)(Def.settings(ss: _*)))
+
+
+// TODO move into sbt-house-rules?
+def cleanupTests(loader: ClassLoader, log: sbt.internal.util.ManagedLogger): Unit = {
+  // shutdown Log4J to avoid classloader leaks
+  try {
+    val logManager = Class.forName("org.apache.logging.log4j.LogManager")
+    logManager.getMethod("shutdown").invoke(null)
```
eed3si9n: Since I am using async logging, I am guessing calling shutdown right away would risk some logs not showing up. But I think it would make sense to remove loggers that are no longer needed at some point.

retronym: Log4J might block until these are written:

```java
    /**
     * Shutdown using the LoggerContext appropriate for the caller of this method.
     * This is equivalent to calling {@code LogManager.shutdown(false)}.
     *
     * This call is synchronous and will block until shut down is complete.
     * This may include flushing pending log events over network connections.
     *
     * @since 2.6
     */
    public static void shutdown() {
        shutdown(false);
    }
```

I haven't tested this.

I don't think it matters much for tests in any case.

eed3si9n: Another problem is that some appenders are long-living (relay appender used by server, backing appender, I think), so I can't stop them. If the loggers are leaking the loaders then I'd need to find out how to prevent that in general, I think.

retronym: To be clear, this is shutting down Log4J that is in the classloader of the tests themselves, not Log4J used in the SBT build that is calling the tests.

If the test were forked, you get away without explicitly shutting down because Log4J registers a shutdown hook to clean up. But that shutdown hook becomes a classloader leak in a "container" environment like a webapp or SBT.

retronym: https://logging.apache.org/log4j/2.x/manual/webapp.html

>  Because of the nature of class loaders within web applications, Log4j resources cannot be cleaned up through normal means. Log4j must be "started" when the web application deploys and "shut down" when the web application undeploys.

eed3si9n: I see. Wouldn't `val logManager = Class.forName("org.apache.logging.log4j.LogManager")` load log4j in the context of the build tho?

retronym: Oh, you're right. I meant to use `loader`. I'll push a fix.

eed3si9n:
I locally still see metaspace climb to 250MB

metaspace2

PR #265, opened 2018-10-03
Scripted integration test infrastructure with Sbt

andreaTP:
This PR contains a basic infrastructure for integration tests of this library with Sbt itself.
I consider this solution strictly better than a bash script since is ways more portable and maintainable.

Here you have the negative test that fail in case of problems with the code (if not cached early).
https://travis-ci.org/andreaTP/librarymanagement/jobs/436603053#L1146
from this branch: https://github.com/andreaTP/librarymanagement/blob/negativeTest

This will make implementing and testing an alternative implementation of `librarymanagement` much easier.

cc. @leonardehrenfried @eed3si9n @alexarchambault

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -259,6 +259,26 @@ lazy val lmIvy = (project in file("ivy"))
     ),
   )
 
+lazy val lmScriptedTest = (project in file("scripted-test"))
+  .enablePlugins(SbtPlugin)
+  .settings(
+    commonSettings,
+    skip in publish := true,
+    name := "scripted-test",
+    scriptedLaunchOpts := { scriptedLaunchOpts.value ++
+      Seq("-Xmx1024M", "-Dplugin.version=" + version.value)
+    },
+    scriptedBufferLog := false
+  ).enablePlugins(SbtScriptedIT)
+
+addCommandAlias("scriptedIvy", Seq(
+  "lmCore/publishLocal",
+  "lmIvy/publishLocal",
+  "lmScriptedTest/clean",
+  """set ThisBuild / scriptedTestLMImpl := "ivy"""",
+  """set ThisBuild / scriptedLaunchOpts += "-Ddependency.resolution=set ThisBuild / dependencyResolution := sbt.librarymanagement.ivy.IvyDependencyResolution(ivyConfiguration.value)" """,
+  "lmScriptedTest/scripted").mkString(";",";",""))
+
 def customCommands: Seq[Setting[_]] = Seq(
   commands += Command.command("release") { state =>
     // "clean" ::
--- a/project/SbtScriptedIT.scala
+++ b/project/SbtScriptedIT.scala
@@ -1,0 +1,106 @@
+import sbt._
+import Keys._
+
+import java.io.File
+import java.util.UUID.randomUUID
+
+object SbtScriptedIT extends AutoPlugin {
+
+  object autoImport {
+    val scriptedTestSbtRepo = settingKey[String]("SBT repository to be used in scripted tests")
+    val scriptedTestSbtRef = settingKey[String]("SBT branch to be used in scripted tests")
+    val scriptedTestLMImpl = settingKey[String]("Librarymanagement implementation to be used in scripted tests")
+    val scriptedSbtVersion = settingKey[String]("SBT version to be published locally for IT tests")
+  }
+
+  import autoImport._
+  override def requires = ScriptedPlugin
+
+  override def trigger = noTrigger
+
+  override lazy val globalSettings = Seq(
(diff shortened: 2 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/265/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

build.sbt
```diff
@@ -259,6 +259,91 @@ lazy val lmIvy = (project in file("ivy"))
...
+      scriptedTests.value
+    }
+  )
+
+addCommandAlias("scriptedIvy", Seq(
+  "lmCore/publishLocal",
+  "lmIvy/publishLocal",
+  "lmScriptedTest/clean",
+  """set ThisBuild / scriptedTestSbtRepo := "https://github.com/andreaTP/sbt"""", // waiting for https://github.com/sbt/sbt/pull/4398 to be merged
+  """set ThisBuild / scriptedTestSbtRef := "b7b94887179053228c88782b316bf0b5973a529a"""", // Last commit in https://github.com/andreaTP/sbt/tree/configurableLibraryManagement
```
muuki88: We would usually set this to `develop` right? 

How would we use this for backports? I imagine that it could be helpful to test things against different versions of SBT. Maybe the `scriptedTestSbtRef` could be a list ?

build.sbt
```diff
@@ -259,6 +259,91 @@ lazy val lmIvy = (project in file("ivy"))
     ),
   )
 
+lazy val scriptedTestSbtRepo = settingKey[String]("SBT repository to be used in scripted tests")
```
muuki88: IMHO this belongs in a separate `AutoPlugin` . It's way to much logic for the `build.sbt`. 😀

muuki88 (commented):
One question and a style nitpick, but apart from that, great work 😍🤗

build.sbt
andreaTP: yes, I do need to have this https://github.com/sbt/sbt/pull/4398 landing in `develop` to use the upstream repo.

uh, that's actually a fortunate side-effect, I didn't think at using this for backports, but it works!
For the moment I would keep it simple with a single `Ref`, if we end up defining too many aliases we can change it later, is it ok for you?

build.sbt
andreaTP: this make sense, give me a little and I do it

andreaTP:
@muuki88 thanks a lot for the review! I think I've addressed your points now, and, yes, an Autoplugin is definitely appropriate here. :smiley:

project/plugins.sbt
```diff
@@ -3,3 +3,5 @@ addSbtPlugin("org.scala-sbt" % "sbt-contraband"  % "0.4.1")
 addSbtPlugin("com.lightbend" % "sbt-whitesource" % "0.1.9")
 
 scalacOptions += "-language:postfixOps"
+
+libraryDependencies += "org.eclipse.jgit" % "org.eclipse.jgit.pgm" % "3.2.0.201312181205-r"
```
eed3si9n: sbt-houserules already contains sbt-git.

project/SbtScriptedIT.scala
```diff
@@ -0,0 +1,106 @@
...
+        IO.write(
+          d / "project" / "build.properties",
+          s"sbt.version=$version"
+        )
+      }
+    }
+  }
+
+  private def sbtUniqueVersion(sbtBaseVersion: String) =
+    s"""$sbtBaseVersion-${randomUUID.toString.replace("-","")}-SNAPSHOT"""
```
eed3si9n: This might end up accumulating too much JAR on cache directories. I think we should hard code to `LM-SNAPSHOT` and also clean it up afterwards on Travis CI.

eed3si9n (changes requested):
Some minor nitpicks, but overall it looks good. Thanks for pushing this effort forward.

project/plugins.sbt
andreaTP: ok, even if I'm not sure to inject in the sbt classpath a specific version of jgit (general observation)

project/SbtScriptedIT.scala
andreaTP: fixed with a `settingKey` and a clean job 'before-cache' in travis, I left commented out an extra clean of `local` since it's not cached after all.

eed3si9n (approved):
Thanks!

PR #280, opened 2018-11-21
Fix CrossVersion.Disabled compat

### LINKED ISSUE #4977: [1.2.x] CrossVersion.Disabled doesn't work
## migration note

`CrossVersion.Disabled` doesn't work as expected in sbt 1.0.x ~ 1.2.x.

We recommend plugin authors to use `CrossVersion.disabled` instead. For pattern matching, you can either check equality with `CrossVersion.disabled` or type match with `case _: sbt.librarymanagement.Disabled`.

## steps 1

```scala
CrossVersion(CrossVersion.Disabled, "1.0.0", "1.0")
```

## problem 1

It doesn't compile.

```scala
<console>:15: error: type mismatch;
 found   : sbt.librarymanagement.Disabled.type
 required: sbt.librarymanagement.CrossVersion
       CrossVersion(CrossVersion.Disabled, "1.0.0", "1.0")
```

## steps 2

```scala
scala> val currentBinaryVersion = "1.0"
currentBinaryVersion: String = 1.0

scala>   def scalaJSMapped(cross: CrossVersion): CrossVersion = cross match {
           case CrossVersion.Disabled =>
             CrossVersion.constant("sjs" + currentBinaryVersion)
           case cross: sbt.librarymanagement.Constant =>
             cross.withValue("sjs" + currentBinaryVersion + "_" + cross.value)
           case cross: CrossVersion.Binary =>
             cross.withPrefix("sjs" + currentBinaryVersion + "_" + cross.prefix)
           case cross: CrossVersion.Full =>
             cross.withPrefix("sjs" + currentBinaryVersion + "_" + cross.prefix)
         }
scalaJSMapped: (cross: sbt.CrossVersion)sbt.CrossVersion

scala> scalaJSMapped(CrossVersion.disabled)
```

From https://github.com/scala-js/scala-js/blob/7c502f61c87735898c2293cb937eee8f55ad8686/sbt-plugin/src/main/scala/org/scalajs/sbtplugin/ScalaJSCrossVersion.scala

## problem 2

```scala
scala> scalaJSMapped(CrossVersion.disabled)
scala.MatchError: Disabled() (of class sbt.librarymanagement.Disabled)
  at .scalaJSMapped(<console>:15)
  ... 36 elided
```

## expectation

```scala
scala> CrossVersion(CrossVersion.Disabled, "1.0.0", "1.0")
res3: Option[String => String] = None

scala> scalaJSMapped(CrossVersion.disabled)
res5: sbt.CrossVersion = Constant(sjs1.0)
```

## notes

This will be fixed in 1.3.0, but it doesn't currently work in 1.2.x.

- sbt/librarymanagement#121 added `final val Disabled = sbt.librarymanagement.Disabled` but it was just a companion object
- (only available in 1.3.0-x) sbt/librarymanagement#280 actually made it `final val Disabled = sbt.librarymanagement.Disabled()`, but this broke Cat's build that was calling `CrossVersion.Disabled()`
- sbt/librarymanagement#290 changed to `final def Disabled = sbt.librarymanagement.Disabled` and `object Disabled extends sbt.librarymanagement.Disabled`, but this caused #4975
- sbt/librarymanagement#316 changes back to `final val Disabled = sbt.librarymanagement.Disabled` (but because we changed the companion object in sbt/librarymanagement#290 that's ok)

dwijnand:
Fixes https://github.com/sbt/sbt/issues/4977

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -157,6 +157,12 @@ lazy val lmCore = (project in file("core"))
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ArtifactExtra.extension"),
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ArtifactTypeFilterExtra.types"),
 
+      // by mistake we aliased the companion object instead of an instance of Disabled
+      // but it was aliased as a constant expression, so even if the binary API has changed
+      // there are no call sites in bytecode because the value got inlined
+      // also it's wouldn't work so I doubt anyone has made use of it
+      exclude[DirectMissingMethodProblem]("sbt.librarymanagement.CrossVersionFunctions.Disabled"),
+
       // contraband issue
       exclude[DirectMissingMethodProblem]("sbt.internal.librarymanagement.ConfigurationReportLite.copy*"),
       exclude[DirectMissingMethodProblem]("sbt.internal.librarymanagement.UpdateReportLite.copy*"),
--- a/core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
+++ b/core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
@@ -8,7 +8,7 @@ final case class ScalaVersion(full: String, binary: String)
 private[librarymanagement] abstract class CrossVersionFunctions {
 
   /** Compatibility with 0.13 */
-  final val Disabled = sbt.librarymanagement.Disabled
+  final val Disabled = sbt.librarymanagement.Disabled()
   final val Binary = sbt.librarymanagement.Binary
   final val Constant = sbt.librarymanagement.Constant
   final val Full = sbt.librarymanagement.Full
@@ -37,7 +37,7 @@ private[librarymanagement] abstract class CrossVersionFunctions {
   def binary: CrossVersion = Binary()
 
   /** Disables cross versioning for a module. */
-  def disabled: CrossVersion = Disabled()
+  def disabled: CrossVersion = Disabled
 
   /** Cross-versions a module with a constant string (typically the binary Scala version).  */
   def constant(value: String): CrossVersion = Constant(value)


dwijnand:
Need to add MiMa exclusions for these.  I doubt anyone's used them seeing how they're wrong, but also they're inlined constants, so even if they're binary API has changed there are no call sites in bytecode because the values got inlined.

dwijnand:
Reported the false positive in MiMa: https://github.com/lightbend/migration-manager/issues/266

dwijnand:
@eed3si9n WDYT?

core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -5,19 +5,23 @@ import sbt.librarymanagement.syntax._
...
-private[librarymanagement] abstract class CrossVersionFunctions {
+private[librarymanagement] trait CrossVersionCompatAliases { self: CrossVersionFunctions =>
 
   /** Compatibility with 0.13 */
-  final val Disabled = sbt.librarymanagement.Disabled
-  final val Binary = sbt.librarymanagement.Binary
-  final val Constant = sbt.librarymanagement.Constant
-  final val Full = sbt.librarymanagement.Full
-  final val Patch = sbt.librarymanagement.Patch
+  final val Disabled = sbt.librarymanagement.Disabled()
```
eed3si9n, line 11: The corresponding construct in sbt 0.13 https://github.com/sbt/sbt-zero-thirteen/blob/v0.13.17/ivy/src/main/scala/sbt/CrossVersion.scala#L19:

```scala
  object Disabled extends CrossVersion { override def toString = "disabled" }
```

so the above change could be considered valid for source compat with sbt 0.13, but in general we probably should recommend migrating to lowercase `CrossVersion.disabled`.

core/src/main/scala/sbt/librarymanagement/CrossVersionExtra.scala
```diff
@@ -5,19 +5,23 @@ import sbt.librarymanagement.syntax._
...
+  final val Binary = sbt.librarymanagement.Binary()
```
eed3si9n: Binary and Full were classes in sbt 0.13, but had lower case value `binary` and `full`.
So keeping `Binary` as the companion object of `sbt.librarymanagement.Binary` kind of makes sense to me.

eed3si9n (changes requested):
Currently the final vals are pointing to the companion objects, and I think we should encourage ppl to use the lowercase versions like `full` and `binary` in sbt 0.13.
The only odd duck is `Disabled`, which was an object in sbt 0.13, so if you want to make this change for better source compat, I think we should limit it to `Disabled`.

Also could we target `develop` branch first?

dwijnand:
Yep `CrossVersion.Disabled` was what lead me to this change, so I'm happy to restrict this PR to just that.  Also, it's needn't be on the bugfix highway, so I'm happy to re-target it to develop.

eed3si9n (approved):
Thanks @dwijnand

eed3si9n:
@dwijnand We might need to revert this PR since @cunei found breakage in the community build:

```
[cats] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
[cats-effect] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
[circe] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
[coursier] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
[scalacheck] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
[scalaz] [error] java.lang.NoSuchMethodError: sbt.librarymanagement.CrossVersion$.Disabled()Lsbt/librarymanagement/Disabled$;
```

dwijnand:
It's a false positive because it's a folded constant.

But MiMa isn't aware of those... https://github.com/lightbend/migration-manager/issues/266

build.sbt
```diff
@@ -157,6 +157,12 @@ lazy val lmCore = (project in file("core"))
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ArtifactExtra.extension"),
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ArtifactTypeFilterExtra.types"),
 
+      // by mistake we aliased the companion object instead of an instance of Disabled
+      // but it was aliased as a constant expression, so even if the binary API has changed
+      // there are no call sites in bytecode because the value got inlined
```
dwijnand, line 162: More precisely "constant folded" not "inlined".

eed3si9n:
@cunei Do you have a stacktrace? I am guessing some plugin out there is actually calling `Disabled()`.

dwijnand:
True, it must be.  Going to ask the Scala team about constant folding.

eed3si9n:
I wonder if we can keep the bincompat by adding `def apply() = this`.

dwijnand:
Yeah, I wondered that too, but there's no `apply` in those java method signatures.

eed3si9n:
ok so `Disabled` is the name of the static method on `CrossVersion$` class.

eed3si9n:
Change `val` to `def`?

eed3si9n:
https://github.com/sbt/librarymanagement/pull/290

Issue #297, opened 2019-04-02
Custom HTTP client


eed3si9n:
There's been a few requests around the HTTP request customization.
It might be convenient to expose the http client.

PR #298, opened 2019-04-03
Add OkHttpClient parameter

eed3si9n:
Ref #297

This adds a parameter so we can pass in OkHttpClient, if someone wants to customize HTTP behavior.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -257,6 +257,11 @@ lazy val lmIvy = (project in file("ivy"))
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ivy.ExternalIvyConfiguration.copy*"),
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ivy.InlineIvyConfiguration.copy*"),
       exclude[DirectMissingMethodProblem]("sbt.librarymanagement.ivy.IvyPaths.copy*"),
+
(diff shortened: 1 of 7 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/298/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eatkins:
Not sure what the dependency story is with library management, but it would it be possible to define a simple interface that mirrors the OkHttpClient api so that we don't add an explicit reference to OkHttpClient in the api?

eed3si9n:
@eatkins Gigahorse is that, an abstraction of HTTP client, but the interesting feature that would make it useful to customize the HTTP client beyond what's there out of the box is in the underlying OkHttpClient (and/or its builder).

eatkins:
This feels almost directly in opposition to https://github.com/sbt/sbt/issues/4597. My concern is that what if someone wants to use a different version of okhttpclient than what lm is using? This would introduce classpath eviction issues.

eatkins:
More generally, my philosophy is that unless a library is designed to directly enhance some other library, it should hide its dependencies.

eed3si9n:
Discussed this during the sbt meeting today. We can probably merge this to LM, but potentially hide the extension points in sbt behind some experimental plugin.

PR #303, opened 2019-05-09
Add toString for DirectCredentials

raboof:
To make 'show credentials' more useful

### DIFF
--- a/ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala
+++ b/ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala
@@ -76,11 +76,13 @@ object Credentials {
 
 sealed trait Credentials
 final class FileCredentials(val path: File) extends Credentials {
-  override def toString = "FileCredentials('" + path + "')"
+  override def toString = s"""FileCredentials("$path")"""
 }
 final class DirectCredentials(
     val realm: String,
     val host: String,
     val userName: String,
     val passwd: String
-) extends Credentials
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/303/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala
```diff
@@ -83,4 +83,5 @@ final class DirectCredentials(
     val host: String,
     val userName: String,
     val passwd: String
+  override def toString = s"DirectCredentials('$realm', '$host', '$userName', ****)"
```
eed3si9n: `val passwd: String` is in the param list, so you need `) extends Credentials { ... }` block.

eed3si9n (changes requested):

ivy/src/main/scala/sbt/librarymanagement/ivy/Credentials.scala
raboof: Doh sorry :/

eed3si9n (approved):
Thanks!

PR #317, opened 2019-08-27
Flip the default to not use Gigahorse by default 

### LINKED ISSUE #3570: sbt 1.0.2 publish with credentials fails to push jar and sources artefacts (for one of three subprojects)
This one is weird, and I likely won't be able to formulate it according to the contribution guidelines. But I think it's worth reporting in order to see whether others are seeing similar cases on-field as well.

In short:

- upgrading one of my sbt projects (multiproject, with three subprojects) from sbt 0.13.15 to 1.0.2 causes one of the projects (always the same) to be only partially published (pom's are there, but jars and sources aren't) to the company's Nexus Maven repo

Screenshots: 

screen shot 2017-09-21 at 11 41 58
**Properly published subproject**

screen shot 2017-09-21 at 11 41 36
**Partial publish (no jars, no sources)**

This happens for both release and snapshot publishes.

All subprojects share the same credentials setup:

```
def publishAs(name: String) = {
  val nexus = "https://maven.myCompany.net/"

  Def.settings(
    moduleName := name,
    publishMavenStyle := true,
    publishArtifact in (Compile, packageDoc) := false,
    publishTo := {
      if (isSnapshot.value) {
        Some("snapshots" at nexus + "content/repositories/snapshots")
      } else {
        Some("releases" at nexus + "content/repositories/releases")
      }
    },
    credentials ++= (for {
      user <- Option(System.getenv().get("MAVEN_USER"))
      pw <- Option(System.getenv().get("MAVEN_PASSWORD"))
    } yield Credentials("Sonatype Nexus Repository Manager", "maven.myCompany.net", user, pw)).toSeq
  )
}
```

All of this works, with 0.13.15 in `project/build.properties`. Changing to 1.0.2 makes publishing consistently fail.

On the command line one sees:

`[error] (testkit/*:publish) java.io.IOException: PUT operation to URL https://maven.myCompany.net/content/repositories/snapshots/fi/myCompany/basement/cloak-akka-http-testkit_2.12/0.0.13-SNAPSHOT/cloak-akka-http-testkit_2.12-0.0.13-SNAPSHOT.jar failed with status code 502: Bad Gateway
`

The Nexus log show `401` (Unauthorized) for the failing parts.

My current work-around is simply to not upgrade to 1.0.2 within the particular project.

eed3si9n:
Fixes sbt/sbt#3570

The primary motivation for bringing in Gigahorse was to speed up the dependency resolution, especially in a high-latency environment like Australia, by enabling keep-alive. One of unintended consequences was that it ended up exposing either a bug in okhttp or Nexus / Artifactory's handling of keep-alive, since corporate users behind these proxy repositories started to experience problem publishing. Multiple people have also reported that putting network proxies like HAProxy works around this issue, which seems consistent with the theory.

Now that dependency resolution has switched to using Coursier by default, I am just going to flip the default here so publishing would use Ivy's default URL handler based on `java.net.HttpURLConnection` - https://github.com/sbt/ivy/blob/5681e1a77a149b93a4cfbe4a86d0263d7bd223eb/src/java/org/apache/ivy/util/url/BasicURLHandler.java

### DIFF
--- a/core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
+++ b/core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
@@ -10,6 +10,7 @@ package internal
 package librarymanagement
 
 import java.util.Locale
+import scala.util.control.NonFatal
 
 object LMSysProp {
   def booleanOpt(name: String): Option[Boolean] =
@@ -40,17 +41,21 @@ object LMSysProp {
 
   val useSecureResolvers: Boolean = getOrTrue("sbt.repository.secure")
 
-  def modifyVersionRange: Boolean = getOrTrue("sbt.modversionrange")
+  lazy val modifyVersionRange: Boolean = getOrTrue("sbt.modversionrange")
 
   lazy val isJavaVersion9Plus: Boolean = javaVersion > 8
   lazy val javaVersion: Int = {
-    // See Oracle section 1.5.3 at:
-    // https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html
-    val version = sys.props("java.specification.version").split("\\.").toList.map(_.toInt)
-    version match {
-      case 1 :: minor :: _ => minor
-      case major :: _      => major
-      case _               => 0
+    try {
+      // See Oracle section 1.5.3 at:
+      // https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html
+      val version = sys.props("java.specification.version").split("\\.").toList.map(_.toInt)
+      version match {
+        case 1 :: minor :: _ => minor
+        case major :: _      => major
+        case _               => 0
+      }
+    } catch {
+      case NonFatal(_) => 0
     }
   }
 
--- a/core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
+++ b/core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
@@ -1,0 +1,58 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
+ * Copyright 2008 - 2010, Mark Harrah
+ * Licensed under Apache License 2.0 (see LICENSE)
+ */
+
+package sbt
+package internal
+package librarymanagement
+
+import java.util.Locale
+
+object LMSysProp {
+  def booleanOpt(name: String): Option[Boolean] =
+    sys.props.get(name).flatMap { x =>
+      x.toLowerCase(Locale.ENGLISH) match {
+        case "1" | "always" | "true" => Some(true)
+        case "0" | "never" | "false" => Some(false)
+        case "auto"                  => None
+        case _                       => None
+      }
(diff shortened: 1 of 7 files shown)


core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
```diff
@@ -0,0 +1,58 @@
...
+
+  val useSecureResolvers: Boolean = getOrTrue("sbt.repository.secure")
+
+  def modifyVersionRange: Boolean = getOrTrue("sbt.modversionrange")
+
+  lazy val isJavaVersion9Plus: Boolean = javaVersion > 8
+  lazy val javaVersion: Int = {
+    // See Oracle section 1.5.3 at:
+    // https://docs.oracle.com/javase/8/docs/technotes/guides/versioning/spec/versioning2.html
+    val version = sys.props("java.specification.version").split("\\.").toList.map(_.toInt)
```
eatkins: This should probably be in a try catch in case the property is somehow null or if there are strings in the java version (unlikely but someone could be testing with an experimental jvm).

core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
```diff
@@ -0,0 +1,58 @@
...
+  // Good: sbt.offline
+  //
+  // Bad:
+  // sbt.disable.interface.classloader.cache
+  // sbt.task.timings.on.shutdown
+  // sbt.skip.version.write -> sbt.genbuildprops=false
+
+  val useSecureResolvers: Boolean = getOrTrue("sbt.repository.secure")
+
+  def modifyVersionRange: Boolean = getOrTrue("sbt.modversionrange")
```
eatkins: Why is this a def but the others are vals?

eatkins (changes requested):
Even though I know that things have worked ok without the try/catch, I've seen this kind of parsing cause some projects to crash (play did something similar in ~run which made it impossible to test a snapshot version of sbt).

Otherwise looks good.

core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
eed3si9n: I guess I can make it a lazy val. The one above is plain `val` because it gets used in resolver initialization on vals.

```scala
  val TypesafeRepositoryRoot = typesafeRepositoryRoot(useSecureResolvers)
```

eatkins (approved):
LGTM

core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * sbt
+ * Copyright 2011 - 2018, Lightbend, Inc.
```
avdv, line 3: Maybe update the year to 2019 here?

eed3si9n: I'll get around to it when I add sbt-header.

PR #318, opened 2019-08-28
Deprecate HTTP resolvers (take 2)

eed3si9n:
Ref https://github.com/sbt/sbt/issues/4905
Ref https://github.com/sbt/librarymanagement/pull/312

Based on the discussion in https://github.com/sbt/sbt/issues/4905, this will exempt localhost, and also allows per-resolver opt-in for HTTP.

To opt-in, I am no longer displaying logs directly from lm, and instead let sbt call a method to perform validation using its logger.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -238,6 +238,9 @@ lazy val lmCore = (project in file("core"))
       exclude[DirectMissingMethodProblem](
         "sbt.librarymanagement.ResolverFunctions.useSecureResolvers"
       ),
+      exclude[ReversedMissingMethodProblem](
+        "sbt.librarymanagement.MavenRepository.allowInsecureProtocol"
+      )
     ),
   )
   .configure(addSbtIO, addSbtUtilLogging, addSbtUtilPosition, addSbtUtilCache)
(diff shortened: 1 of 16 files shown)


eatkins:
LGTM.

henricook:
This really needs some docs - I include all my resolvers in `~/.sbt/repositories` in my CI setup. They're all local dependency proxies and don't use HTTPS between CI and the proxy (proxy uses HTTPS to the real internet)

I don't think there's a way for me to stop getting these errors at the moment, with my use case?

Openpalm:
@henricook got a solution to the issue yet ?

henricook:
Unfortunately not, some further work was suggested on https://github.com/sbt/sbt/issues/4905 but I don't think anything has happened yet.

Openpalm:
comment: tried to actually switch http to https under [repositiries] but that caused typhoon of SSL handshae errors due to the certs not being in my truststore.
i know this can be fixed via manual import of certs. but feels like double work. waiting for a fix for the main push here, and marking this a "will do when we need to" thing project wise. for now things work.

i'll revert to 1.2.8 for now, since it's honestly good enough and it will be a while before we move to 2.13.0

hopefully by then there will be a solution available.

PR #329, opened 2020-03-25
UpdateOptions: make latestSnapshots = false the default

smarter:
I'm not aware of a single person who enjoys this feature. In fact, it is
so cursed that sbt warns you every time you accidentally use it (and
turning it off properly is non-trivial: setting it in the current
project is not enough, you'll need to set it in the meta-project,
meta-meta-project, etc).

Let's do the right thing for our users by giving them sensible defaults
instead.

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/ivyint/SbtChainResolver.scala
@@ -225,9 +225,6 @@ private[sbt] case class SbtChainResolver(
 
         if (resolvedModule.getId.getRevision.contains("SNAPSHOT")) {
 
-          Message.warn(
-            "Resolving a snapshot version. It's going to be slow unless you use `updateOptions := updateOptions.value.withLatestSnapshots(false)` options."
-          )
           val resolvers = sortedRevisions.map(_._2.getName)
           sortedRevisions.foreach(h => {
             val (module, resolver) = h
--- a/ivy/src/main/scala/sbt/librarymanagement/ivy/UpdateOptions.scala
+++ b/ivy/src/main/scala/sbt/librarymanagement/ivy/UpdateOptions.scala
@@ -108,7 +108,7 @@ object UpdateOptions {
     new UpdateOptions(
       circularDependencyLevel = CircularDependencyLevel.Warn,
       interProjectFirst = true,
-      latestSnapshots = true,
+      latestSnapshots = false,
       cachedResolution = false,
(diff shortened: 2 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/329/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n:
> Let's do the right thing for our users by giving them sensible defaults instead.

Normally people use `-SNAPSHOT` because they want the most recently published/built artifacts, either from `publishLocal` or from the corporate proxy. The current default value is the safe and more sensible default from that perspective. If you set the latestSnapshot to `false`, you could end up with SNAPSHOT that's old. Lots of people complained about this since 2012 (https://github.com/sbt/sbt/issues/321) until we fixed it in https://github.com/sbt/sbt/pull/1520.

smarter:
If we're resolving locally then it's not slow and we shouldn't bet getting a warning. If people are using this through a corporate proxy then I'm very surprised that they're not complaining about the constant warnings that cannot be suppressed.

dwijnand:
So it still needs to check all the resolvers for who has "latest" (then it's fast to retrieve from the local dir), but I guess with this modern Coursier technology  that initial check is now fast.

So maybe the best change to do is to drop the warning but keep the semantics?

eed3si9n:
> So maybe the best change to do is to drop the warning but keep the semantics?

Technically this shouldn't happen if the resolvers are not used, maybe? If we are still seeing this then Ivy is somehow being invoked either by publish task or by Dotty plugin?

dwijnand:
Ah, right, this is the Ivy code path.  Then, yeah, the current default I think is still the better default: semantically pick the latest, but inform the user that the use of SNAPSHOTs incurs a performance penalty.

smarter:
So what happens exactly in the coursier code path by default ? Does it completey ignore latestSnapshots and always acts as if latestSnapshots = false ?

eed3si9n:
AFAIK none of the UpdateOptions are respected. On the other hand Coursier puts in place a more reasonable caching policy for snapshots. I think it checks the timestamp for the local files, and cache for TTL (24h) if it's remote.
https://get-coursier.io/docs/cache

smarter:
OK, I guess we can close then, but UpdateOptions documentation should warn that it's a legacy thing.

PR #337, opened 2020-06-26
support UNC path

### LINKED ISSUE #5596: Resolver.file mangles UNC paths
If on Windows we do something like this:

````
sbt.Resolver.file("name", new File("//intra/live/foo"))(Patterns("[organisation]/[module]/[revision]/[artifact]-[revision](-[classifier]).[ext]"))
````

Then the resulting `FileResolver` is built with paths like `"C:\intra\live\foo\[organisation]/[module]/[revision]/[artifact]-[revision](-[classifier]).[ext]"` i.e. the UNC path is replaced with the drive letter of the current working directory, which is definitely not what we want.

The cause of the problem is https://github.com/sbt/librarymanagement/blob/c3dca6880d7b129f708c8b59cf703e52fb4554e3/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L307

i.e.
````
new File(baseDirectory.toURI.normalize).getAbsolutePath
````

Because:
````
new File("//intra/live/foo").toURI() == new java.net.URI("file:////intra/live/foo")
new java.net.URI("file:////intra/live/foo").normalize() == new java.net.URI("file:/intra/live/foo")
new File(new java.net.URI("file:/intra/live/foo")) == new File("\\intra\\live\\foo")
new File("\\intra\\live\\foo").getAbsolutePath() == "C:\\intra\\live\\foo"
````

I think you can fix this by just dropping the use of `normalize` since `new File(new File("//intra/live/foo").toURI()).getAbsolutePath()` == `"//intra/live/foo"`

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/5596

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
+++ b/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
@@ -4,7 +4,7 @@
 package sbt.librarymanagement
 
 import java.io.{ IOException, File }
-import java.net.URL
+import java.net.{ URI, URL }
 import scala.xml.XML
 import org.xml.sax.SAXParseException
 import sbt.util.Logger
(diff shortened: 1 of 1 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/337/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #340, opened 2020-09-13
Retry with backoff while publishing

izharahmd:
Refer sbt/sbt#5698

add retry with exponential backoff on transient network failure while publishing artifacts.

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
@@ -25,6 +25,7 @@ import sbt.librarymanagement.{ ModuleDescriptorConfiguration => InlineConfigurat
 import syntax._
 import InternalDefaults._
 import UpdateClassifiersUtil._
+import sbt.internal.librarymanagement.IvyUtil.TransientNetworkException
 
 object IvyActions {
 
@@ -497,8 +498,14 @@ object IvyActions {
       checkFilesPresent(artifacts)
       try {
         resolver.beginPublishTransaction(module.getModuleRevisionId(), overwrite);
-        for ((artifact, file) <- artifacts)
-          resolver.publish(artifact, file, overwrite)
+        artifacts.foreach {
+          case (artifact, file) =>
+            IvyUtil.retryWithBackoff(
+              resolver.publish(artifact, file, overwrite),
+              TransientNetworkException.apply,
+              maxAttempts = LMSysProp.maxPublishAttempts
+            )
+        }
         resolver.commitPublishTransaction()
       } catch {
         case e: Throwable =>
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
@@ -1,6 +1,50 @@
 package sbt.internal.librarymanagement
 
+import java.io.IOException
+import java.net.{ SocketException, SocketTimeoutException }
+
+import scala.annotation.tailrec
+import scala.util.{ Failure, Success, Try }
+
 private[sbt] object IvyUtil {
   def separate[A, B](l: Seq[Either[A, B]]): (Seq[A], Seq[B]) =
     (l.flatMap(_.left.toOption), l.flatMap(_.right.toOption))
+
+  @tailrec
+  final def retryWithBackoff[T](
+      f: => T,
+      predicate: Throwable => Boolean,
+      maxAttempts: Int,
+      retry: Int = 0
+  ): T = {
+    // Using Try helps in catching NonFatal exceptions only
+    Try {
+      f
+    } match {
+      case Success(value)                                          => value
+      case Failure(e) if predicate(e) && retry < (maxAttempts - 1) =>
+        // max 8s backoff
+        val backoff = math.min(math.pow(2d, retry.toDouble).toLong * 1000L, 8000L)
+        Thread.sleep(backoff)
+        retryWithBackoff(f, predicate, maxAttempts, retry + 1)
+      case Failure(e) => throw e
+    }
+  }
+
+  object TransientNetworkException {
+    private val _r = """.*HTTP response code: (503|429).*""".r
+
+    @inline private def check(s: String): Boolean = {
+      if (s == null) return false
+
+      _r.pattern.matcher(s).matches()
+    }
+
+    def apply(t: Throwable): Boolean = t match {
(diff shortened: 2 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/340/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

izharahmd:
@eed3si9n could you please review this?

eed3si9n:
@izharahmd Thanks for the contribution!

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
```diff
@@ -497,8 +498,15 @@ object IvyActions {
...
-        for ((artifact, file) <- artifacts)
-          resolver.publish(artifact, file, overwrite)
+        artifacts.foreach {
+          case (artifact, file) =>
+            // should read max retries from some config?
+            IvyUtil.retryWithBackoff(
+              resolver.publish(artifact, file, overwrite),
+              TransientNetworkException.apply,
+              retryLimit = 5
+            )
```
eed3si9n: https://github.com/sbt/librarymanagement/blob/develop/core/src/main/scala/sbt/internal/librarymanagement/LMSysProp.scala is a place to keep system properties.
You could make something like

```scala
val maxPublishAttempts = java.lang.Integer.getInteger("sbt.repository.publish.retry", 3)
```

ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
```diff
@@ -1,6 +1,50 @@
...
+      retry: Int = 0
+  ): T = {
+    // Using Try helps in catching only NonFatal exceptions only
+    Try {
+      f
+    } match {
+      case Success(value)                                         => value
+      case Failure(e) if predicate(e) && (retry < retryLimit - 1) =>
+        // max 42 sec backoff
+        val backoff = math.min(math.pow(2d, retry.toDouble).toLong * 1000L, 42000L)
```
eed3si9n: For our use case I'd suggest we set the max backoff to 8s.

eed3si9n:
I don't know how often this would be hit, but following Gradle, I'd suggest we start with max attempt of 3 including the initial attempt. This is partly to avoid overwhelming the repo under pressure, but also to make sure it doesn't look like sbt is frozen (to human or to Jenkins).

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
izharahmd: Since we are counting attempts instead of retries, should we use property `sbt.repository.publish.attempts`?

ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
```diff
@@ -1,6 +1,50 @@
...
+      if (s == null) return false
+
+      _r.pattern.matcher(s).matches()
+    }
+
+    def apply(t: Throwable): Boolean = t match {
+      case _: SocketException | _: SocketTimeoutException => true
+      case e: IOException if check(e.getMessage)          => true
+      case _                                              => false
+    }
```
izharahmd, line 47: @eed3si9n is there some better way we could identify network errors? This is how gradle does it 
https://github.com/gradle/gradle/blob/master/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/transport/NetworkOperationBackOffAndRetry.java#L58
https://github.com/gradle/gradle/blob/master/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/transport/NetworkingIssueVerifier.java#L27

ivy/src/main/scala/sbt/internal/librarymanagement/IvyActions.scala
eed3si9n: Yea. That sounds good.

ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
eed3si9n: I'm not sure. Whatever you got here seems like a good starting point. We can iterate during patch releases if we need adjustments.

eed3si9n (approved):

ivy/src/main/scala/sbt/internal/librarymanagement/IvyUtil.scala
```diff
@@ -1,6 +1,50 @@
...
+        // max 8s backoff
+        val backoff = math.min(math.pow(2d, retry.toDouble).toLong * 1000L, 8000L)
+        Thread.sleep(backoff)
+        retryWithBackoff(f, predicate, maxAttempts, retry + 1)
+      case Failure(e) => throw e
+    }
+  }
+
+  object TransientNetworkException {
+    private val _r = """.*HTTP response code: (503|429).*""".r
```
rdesgroppes, line 35: What about extending the list to:
1. REQUEST_TIMEOUT = 408
2. BAD_GATEWAY = 502
3. GATEWAY_TIMEOUT = 504

eed3si9n: Looking at Graldle's code it seems to extend this to all 5XX according to the comment - https://github.com/gradle/gradle/blob/master/subprojects/dependency-management/src/main/java/org/gradle/api/internal/artifacts/repositories/transport/NetworkingIssueVerifier.java#L27

izharahmd: should we implement this? Regex would look something like like: `.*HTTP response code: (5\d{2}|408|429).*`

eed3si9n: Sure why not?

rdesgroppes: @izharahmd will you do it or should I?

izharahmd: @rdesgroppes I have drafted the PR. Adding test cases which i missed in the last PR :)

PR #349, opened 2020-12-11
Add Ordering For VersionNumber

isomarcte:
This commit adds an `Ordering` to the companion object for `VersionNumber`. This is nice in many context, e.g. working with MiMa, writing plugins, etc.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
+++ b/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
@@ -260,9 +260,9 @@ object VersionNumber {
    *
    * This ordering compares a version number first by comparing the
    * numbers. Numbers with less components are ordered < longer ones,
-   * e.g. `"1.0" < "1.0.0" == true`. If the [[VersionNumber#numbers]] is
-   * equal then [[VersionNumber#tags]] is compared. If [[VersionNumber#tags]]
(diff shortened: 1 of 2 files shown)


core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
```diff
@@ -255,6 +255,43 @@ object VersionNumber {
...
+              case 0 =>
+                x.extras.zipAll(y.extras, "", "").foldLeft(0)(foldFunction[String])
+              case otherwise =>
+                otherwise
+            }
+          case otherwise =>
+            otherwise
+        }
+    }
+  }
```
eed3si9n, line 294: I'm actually against implementing ordering for `VersionNumber` here. Not for lack of want for the feature, but because we shouldn't do a half-baked solution to a complicated problem. Apache Ivy and Coursier are the implementation that would do ordering. The version numbers look simple at first, but there are a bunch of corner cases like `+metadata`, `1.0-alpha`, `1.0-RC1`, `1.0-MX`, `1.0a` to name a few. Not that Coursier is perfect, but anyone trying to sort the version number should use its sorting for consistency. See also https://github.com/coursier/coursier/pull/1315

isomarcte:
@eed3si9n :+1: I'll close this.

PR #352, opened 2020-12-19
Fixes match error when using withDottyCompat

eed3si9n:
Fixes https://github.com/sbt/sbt/issues/6210

scodec-bits is published with pvp `versionScheme` (nice), this means that we should just evaluate the version portion for pvp-ness, but I was using `guessSecondSegment` that checks for Scala suffix. That's mistake 1.

`guessSecondSegment` assumes that the Scala suffix uses the given ScalaModuleInfo, but with 2.13-3 sandwich we can't assume this. In the reported case, Scala module is 3.0.0-M3 but scodec-bits uses 2.13. So that's mistake 2.

This attempts to correct both the mistakes.
1. Instead of `guessSecondSegment`, this adds a simpler `evalPvp` function.
2. `guessSecondSegment` just looks for `_2.` or `_3` and ignores the Scala module.

PR #353, opened 2020-12-19
[1.4.x] Fixes match error when using withDottyCompat

eed3si9n:
This is a backport of https://github.com/sbt/librarymanagement/pull/352
Fixes https://github.com/sbt/sbt/issues/6210

scodec-bits is published with pvp `versionScheme` (nice), this means that we should just evaluate the version portion for pvp-ness, but I was using `guessSecondSegment` that checks for Scala suffix. That's mistake 1.

`guessSecondSegment` assumes that the Scala suffix uses the given ScalaModuleInfo, but with 2.13-3 sandwich we can't assume this. In the reported case, Scala module is 3.0.0-M3 but scodec-bits uses 2.13. So that's mistake 2.

This attempts to correct both the mistakes.
1. Instead of `guessSecondSegment`, this adds a simpler `evalPvp` function.
2. `guessSecondSegment` just looks for `_2.` or `_3` and ignores the Scala module.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
+++ b/core/src/main/scala/sbt/librarymanagement/EvictionWarning.scala
@@ -98,12 +98,25 @@ object EvictionWarningOptions {
   lazy val defaultGuess: Function1[(ModuleID, Option[ModuleID], Option[ScalaModuleInfo]), Boolean] =
     guessSbtOne orElse guessSecondSegment orElse guessSemVer orElse guessFalse
 
+  private def isNameScalaSuffixed(name: String): Boolean =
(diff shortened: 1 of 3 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=1.4.x&zinc=1.4.x&io=1.4.x&librarymanagement=pull/353/head&util=1.4.x&website=1.4.x&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #356, opened 2020-12-21
EvictionError

eed3si9n:
Ref https://eed3si9n.com/enforcing-semver-with-sbt-strict-update

This adds EvictionError, a replacement for EvictionWarning. The problem with the current eviction warning is that it has too many false positives. Using the versionScheme information that could be supplied by the library authors and/or the build users, this eliminates the guessing work. At which point, we can fail the resolution.


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/356/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

PR #359, opened 2020-12-23
Add scala3doc artifact

adpi2:
Makes scala3doc available in Scala 3 tool dependencies.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/ScalaArtifacts.scala
+++ b/core/src/main/scala/sbt/librarymanagement/ScalaArtifacts.scala
@@ -14,11 +14,13 @@ object ScalaArtifacts {
   final val Scala3InterfacesID = "scala3-interfaces"
   final val TastyCoreID = "tasty-core"
   final val Scala3DocID = "scala3doc"
+  final val Scala3TastyInspectorID = "scala3-tasty-inspector"
 
   private[sbt] final val Scala3LibraryPrefix = Scala3LibraryID + "_"
   private[sbt] final val Scala3CompilerPrefix = Scala3CompilerID + "_"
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/359/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

core/src/main/scala/sbt/librarymanagement/ScalaArtifacts.scala
```diff
@@ -13,17 +13,22 @@ object ScalaArtifacts {
   final val Scala3CompilerID = "scala3-compiler"
   final val Scala3InterfacesID = "scala3-interfaces"
   final val TastyCoreID = "tasty-core"
+  final val Scala3DocID = "scala3doc"
```
eed3si9n, line 16: Listing this out makes it clear that scala3doc stands out as as oddly named duck.
@romanowski Do you think we can name it `scala3-doc`? - https://github.com/lampepfl/scala3doc/issues/264

eed3si9n (approved):

core/src/main/scala/sbt/librarymanagement/ScalaArtifacts.scala
romanowski: Maybe this may make sense. We should probably decide between `scaladoc` and `scala3-doc` then. cc @abgruszecki

PR #370, opened 2021-03-09
Specify dependencies of plugins that handles custom protocols

MasseGuillaume:
In coursier, you can handle custom protocols by defining a protocol handler:
https://get-coursier.io/docs/2.0.10/extra.html#extra-protocols
https://github.com/coursier/coursier/blob/ce251d0288bf029e4eeb8700e8990178559a005d/modules/tests/jvm/src/test/scala/coursier/cache/protocol/TestprotocolHandler.scala

I'm currently implementing support for Google Aritfactr Resitory:
https://github.com/coursier/coursier/issues/1987#issuecomment-792894989

When dealing with artifact with protocol `artifactregistry` I need to have a plugin available in the classpath. Since lmcoursier is in the sbt launcher classpath, it's not available by adding `libraryDependencies += ...` in `project/plugins.sbt` like you would do for an sbt plugin. Therefore, I propose a way to explicitly declare the artifacts you need to handle the protocol handling.

### DIFF
--- a/core/src/main/contraband-scala/sbt/librarymanagement/LibraryManagementCodec.scala
+++ b/core/src/main/contraband-scala/sbt/librarymanagement/LibraryManagementCodec.scala
@@ -10,7 +10,6 @@ trait LibraryManagementCodec extends sjsonnew.BasicJsonProtocol
   with sbt.librarymanagement.UpdateLoggingFormats
   with sbt.internal.librarymanagement.formats.LogicalClockFormats
   with sbt.librarymanagement.ArtifactTypeFilterFormats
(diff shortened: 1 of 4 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/370/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

MasseGuillaume:
I found another way via: https://github.com/sbt/sbt/pull/6375

PR #377, opened 2021-05-14
Fixes sbt/sbt#6496 - bug in version parser

Nirvikalpa108:
## What is the problem?
sbt/sbt#6496 identifies a bug in the logic which assesses whether different versions of the same transitive dependency are binary compatible. 

If one of the transitive dependencies included a version with a full-stop literal, it would be parsed incorrectly and an error would be thrown to the user falsely saying that the dependencies are binary incompatible.

## What is the solution?
This PR fixes the regex used by the parser to include full-stop literals.

## Is there anything else to be done?
It is worth us checking the rest of the codebase in case this bug might exist in other parsing functions. The tests are super helpful for figuring out which strings might break the current logic 👍🏽.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
+++ b/core/src/main/scala/sbt/librarymanagement/VersionNumber.scala
@@ -66,7 +66,7 @@ object VersionNumber {
     def splitDash(s: String) = splitOn(s, '-')
     def splitPlus(s: String) = splitOn(s, '+') map ("+" + _)
 
-    val TaggedVersion = """(\d{1,14})([\.\d{1,14}]*)((?:-\w+)*)((?:\+.+)*)""".r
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/377/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

eed3si9n (approved):
Thanks Amina!

eed3si9n:
> It is worth us checking the rest of the codebase in case this bug might exist in other parsing functions. 

Here's my follow-up on similar version parsing code - https://github.com/sbt/librarymanagement/pull/378

PR #380, opened 2021-05-28
Expand allowed maven repository id characters to match maven

### LINKED ISSUE #1608: MakePom isValidIDCharacter method is overly restrictive
I found very little information on what are supposed to be valid characters for repository id. Maven documentation says lowercase letters and "no strange symbols", but does not get into any more detail than that. SBT keeps it to letters and numbers, which leaves out at least the dash (-), a character that is widely used on examples throughout Maven documentation.

As it happens, maven itself will complain if it sees "strange symbols", producing the following warning:

```
[WARNING] 'repositories.repository.id' must not contain any of these characters \/:"<>|?* but found / @ line 74, column 11
```

Interestingly, it doesn't mention space, which also seems to be illegal. And I'd be wary of using parenthesis.

I think this method should at least allow dash, as it is widely used in examples. There is some consequence to this -- on internal networks were all repository access goes to a password-protected proxy, the id of the repository on an artifact is essential for maven to find the proper security setting.

peter-janssen:
fixes sbt/sbt#1608

see https://github.com/apache/maven/blob/5c45b3fe22d8ec4270b2cf1016581b8927d2913d/maven-model-builder/src/main/java/org/apache/maven/model/validation/DefaultModelValidator.java#L84

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
@@ -504,7 +504,7 @@ class MakePom(val log: Logger) {
     s.toArray.map(_.asInstanceOf[DependencyResolver])
 
   def toID(name: String) = checkID(name.filter(isValidIDCharacter).mkString, name)
-  def isValidIDCharacter(c: Char) = """\/:"<>|?*""".contains(c)
+  def isValidIDCharacter(c: Char) = !"""\/:"<>|?*""".contains(c)
   private def checkID(id: String, name: String) =
     if (id.isEmpty) sys.error("Could not convert '" + name + "' to an ID") else id
   def mavenRepository(repo: MavenRepository): XNode =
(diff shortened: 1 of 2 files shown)


typesafe-tools:
To validate this pull request against other sbt modules, please visit <A HREF="https://jenkins.scala-sbt.org:8592/job/sbt-validator/parambuild/?sbt=develop&zinc=develop&io=develop&librarymanagement=pull/380/head&util=develop&website=develop&deploy_to_bintray=false&run_sbt_community=false&verbose_dbuild=false&community_verbose=false">this link.</A>

ivy/src/main/scala/sbt/internal/librarymanagement/MakePom.scala
```diff
@@ -504,7 +504,7 @@ class MakePom(val log: Logger) {
     s.toArray.map(_.asInstanceOf[DependencyResolver])
 
   def toID(name: String) = checkID(name.filter(isValidIDCharacter).mkString, name)
-  def isValidIDCharacter(c: Char) = c.isLetterOrDigit
+  def isValidIDCharacter(c: Char) = """\/:"<>|?*""".contains(c)
```
eed3si9n: I wonder if the logic is flipped here?

eed3si9n:
Thanks for the contribution. If there is a convenient place to hang it, maybe there should be a test to see if `-` would now be admitted.

peter-janssen:
I've added a test but I was unsure as to where to put is.

eed3si9n (approved):
Thanks!

PR #382, opened 2021-07-04
[2.x] Replace tuple with proper record type for licenses

mdedetrich:
So when cleaning up some Scala sbt projects that was using `xml` literal syntax to add POM info I noticed that some projects (i.e. https://github.com/getquill/quill/pull/2187/files) use the `distribution` and `comments` fields for the `<license>` section (see https://maven.apache.org/pom.html#licenses for more info).

Unfortunately it doesn't appear to be possible to set this information using the `licenses` setting because its type is set to `Tuple[(String, URL)]`, it being a tuple makes it impossible to add extra fields without breaking the type information (note that for this reason and others, its generally discouraged to use `Tuple` for public return types in API's because they are impossible to extend)

This PR replaces `licenses` with a proper record type that is auto generated with contraband (in similar fashion to `SCMInfo` and others). There are however some interest points to note

1. When it comes to integrating this into https://github.com/sbt/sbt, even though this change is technically a breaking change its theoretically possible to add a specific overloaded `:=`, `++=`, `+=` that just works on the specific `Seq[(String, URL)]` type so that current
    ```scala
    licenses += ("Apache-2.0", url("https://opensource.org/licenses/Apache-2.0"))
    ```
    Can still work. Thankfully there is no other setting that has type `[Seq[(String, URL)]` although this is definitely a case of a workaround/hack and for this reason I would suppose using the `@deprecated` annotation would be recommended so that in the following major bump of SBT people will do
    ```scala
    licenses += LicenseInfo("Apache-2.0", url("https://opensource.org/licenses/Apache-2.0"))
    ```
   Instead
2. Interestingly Ivy only uses 2 fields for License, `name` and `url` (this is evident in both the Ivy api that librarymanagement uses and the documentation at https://ant.apache.org/ivy/history/2.4.0/ivyfile/license.html). A point can be made though that you can be using SBT only for maven (i.e. you want to use SBT instead of maven for Java projects) in which case those 2 extra fields are definitely useful (albeit I haven't used these extra fields to populate the pom info currently in this PR).

Because of point 2 it can be argued that this change is pointless since currently Ivy only really uses the `name` and `url` field however at a bare minimum this at least fixes an obvious code smell of having `Tuple` in a public API and solving this sooner rather than later is always ideal. As a longshot it can also be said that in the future SBT may use an additional distribution method that is not just Ivy, in which case this can be a limitation.

## Notes
I have used `1.6.0` as the version for the latest breaking change. I do not know if this version corresponds to the main sbt release or the release of librarymanagement specifically. Also I am not sure if technically speaking **all** of the fields of the newly created `LicenseInfo` need to be marked with `since` (even though in all previous versions of SBT you still supplied those fields, it was just in the form of a `Tuple`)


ivy/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
```diff
@@ -313,7 +313,7 @@ final class IvySbt(
         mod.setHomePage(h.toString)
       }
       moduleInfo.licenses foreach { l =>
-        mod.addLicense(new License(l._1, l._2.toString))
+        mod.addLicense(new License(l.name, l.url.toString))
```
mdedetrich, line 335: As can be seen here, the Ivy API hardcoded license to only contain 2 fields however we can also add the extra `distribution`/`comments` field to the raw POM manifest itself since it is part of maven.

I haven't done this currently but if people agree with the direction of the PR I can add this functionality in there.

mdedetrich:
So as an update, I just published this PR locally so I could use it with sbt and in regards to the `SettingsKey` `Structure` the following works

```scala
  implicit class SettingKeyLicenseDepreciations(settingsKey: SettingKey[Seq[LicenseInfo]]) {
    @deprecated(
      "Use := Seq(LicenseInfo(String, java.net.URL)*) instead of := Seq(String, java.net.URL)*",
      "1.6.0"
    )
    final def :=(v: Seq[(String, java.net.URL)]): Def.Setting[Seq[LicenseInfo]] =
      settingsKey := v.map { case (name, url) => LicenseInfo(name, url) }

    @deprecated(
      "Use += LicenseInfo(String, java.net.URL) instead of += (String, java.net.URL)",
      "1.6.0"
    )
    final def +=[U](v: (String, java.net.URL))(
        implicit a: Append.Value[Seq[LicenseInfo], LicenseInfo]
    ): Def.Setting[Seq[LicenseInfo]] =
      (settingsKey += LicenseInfo(v._1, v._2))

    @deprecated(
      "Use ++= Seq(LicenseInfo(String, java.net.URL)*) instead of ++= Seq((String, java.net.URL)*)",
      "1.6.0"
    )
    final def ++=[U](vs: Seq[(String, java.net.URL)])(
        implicit a: Append.Values[Seq[LicenseInfo], Seq[LicenseInfo]]
    ): Def.Setting[Seq[LicenseInfo]] =
      (settingsKey ++= vs.map { case (name, url) => LicenseInfo(name, url) })

    @deprecated(
      "Use -= LicenseInfo(String, java.net.URL) instead of -= (String, java.net.URL)",
      "1.6.0"
    )
    final def -=[U](
        v: (String, java.net.URL)
    )(implicit r: Remove.Value[Seq[LicenseInfo], LicenseInfo]): Def.Setting[Seq[LicenseInfo]] =
      (settingsKey -= LicenseInfo(v._1, v._2))

    @deprecated(
      "Use --= Seq(LicenseInfo(String, java.net.URL)*) instead of --= Seq((String, java.net.URL)*)",
      "1.6.0"
    )
    final def --=[U](vs: Seq[(String, java.net.URL)])(
        implicit r: Remove.Values[Seq[LicenseInfo], Seq[LicenseInfo]]
    ): Def.Setting[Seq[LicenseInfo]] =
      (settingsKey --= vs.map { case (name, url) => LicenseInfo(name, url) })
  }
  ```
(there is a just a question of where is the best place to put this, I put it in inside of `private[sbt] trait BuildSyntax` but I am not that familiar with sbt and there is likely a better spot for this).

Of course the only other problem is if other SBT plugins happen to use the `license` key. No idea comes to mind on how to solve this problem (I personally think its impossible?) but I suspect very few sbt-plugins actually directly use this value so maybe its a breakage thats worth it?

**Comment by eed3si9n** [comment]
I don't think we can do this until sbt 2.0, which I think doesn't have to be in a distant future now that Scala 2.12 is two versions behind.

**Comment by mdedetrich** [comment]
Understood, if we do this for SBT 2.0 do we need still need something like the above `SettingKeyLicenseDepreciations` as depreciation warnings or can we just do a hard cutoff with along with a scalafix?

I will update the PR to change the `since` version to 2.0.0 and leave it open until we have a 2.0.0 branch (or `develop` becomes `2.0.0`).

Feel free to ping the PR when you want me to ultimately reabse it and/or change the base branch

**Comment by mdedetrich** [comment]
@eed3si9n I have just updated the PR to make the version since 2.0.0 (since this is the first release such a change would be accepted).

Do you think its worth it to also add the `distribution`/`comments` fields to the POM xml license section from the `LicenseInfo`? I had a brief look at the code and from initial glance this may be a bit of tricky since Ivy API itself has no concept of these fields which means that I have to hotpatch the fields afterwards (hopefully this is easy with Scala's XML literals, haven't done this in a while)

**Comment by eed3si9n** [comment]
> Do you think it's worth it to also add the `distribution/comments` fields to the POM xml license section from the `LicenseInfo`?

As far as the publishing using Ivy is concerned, I would say we should keep whatever it's doing as-is.
For the future, we should we pave the way for Ivy-less publishing.

**Comment by mdedetrich** [comment]
@eed3si9n Since 2.0.x work is starting to materialize I have rebased this PR off of latest develop to see that its compatible

File: core/src/main/contraband/librarymanagement.json [hunk 2]
```diff
@@ -726,6 +726,19 @@
...
     },
+    {
+      "name": "LicenseInfo",
+      "namespace": "sbt.librarymanagement",
+      "target": "Scala",
+      "type": "record",
+      "doc": "Basic license information for a project module",
+      "fields": [
+        { "name": "name",          "type": "String",      "default": "\"\"",                        "since": "2.0.0" },
+        { "name": "url",           "type": "java.net.URL","default": "new java.net.URL(\"http:\")", "since": "2.0.0" },
```
mdedetrich: So contraband has changed its behaviour since I opened this PR and its now asking for default values on fields. While for `String` its trivial, with `java.net.URL` doing `new URL("http:")` seems to be the minimal way I can construct the type

jtjeferreira:
@mdedetrich wanna retarget this PR to https://github.com/sbt/librarymanagement/tree/wip/sbt-2.x branch?

mdedetrich:
@jtjeferreira Many thanks for the reminder. @eed3si9n The PR is now ready, shall it be merged into the 2.0 branch?

core/src/main/contraband/librarymanagement.json
```diff
@@ -733,6 +733,19 @@
...
```
jtjeferreira: please use `java.net.URI` here as done in  #443 
```suggestion
        { "name": "url",           "type": "java.net.URI","default": "new java.net.URL(\"http:\")", "since": "2.0.0" },
```

core/src/main/contraband/librarymanagement.json
jtjeferreira: did you retry not setting `since`? Maybe it does not require setting the `default` field if `since` is not set...

core/src/main/contraband/librarymanagement.json
mdedetrich: Changed and pushed.

mdedetrich:
@jtjeferreira I ended up fixing both the default value (by removal of `@since`) as well as changing to `java.net.URI`.

One thing that I did notice is that the contraband generator seems to currently be commented out, so I had to temporarily uncomment it so that I could generate new sources and there seems to be some differences compared to last time.

For example it regenerated `SemComparator` but the new version is only different in terms of whitespace.

In any case I would double check the diff to make sure nothing unintended has been introduced with `generateContrabands`

EDIT: Wrt contraband being disabled I just noticed https://github.com/sbt/librarymanagement/pull/454 , maybe its wiser to merge that PR first and then I can re-run contraband on this PR?

mzuehlke:
I am not sure if https://github.com/sbt/sbt/pull/7739 will make this repo obsolete for sbt 2.x ? @adpi2

adpi2:
> I am not sure if https://github.com/sbt/sbt/pull/7739 will make this repo obsolete for sbt 2.x ?

Yes that's what I am trying to do. But it is not done yet, so you can continue working on here, and I'll merge the changes in the main sbt repo later.

mdedetrich:
I think it still makes sense to merge https://github.com/sbt/librarymanagement/pull/454 before this PR as this PR is also doing contraband schema related changes, no?

adpi2:
> I think it still makes sense to merge https://github.com/sbt/librarymanagement/pull/454 before this PR as this PR is also doing contraband schema related changes, no?

Done. @mdedetrich Could you fix the conflict? :pray:

mdedetrich:
@adpi2 I just rebased, fixed conflicts and re-ran `generateContrabands` and but it did create some suspicious diffs (i.e. check `ivy/src/main/contraband-scala/sbt/librarymanagement/ivy/ExternalIvyConfigurationFormats.scala`)

jtjeferreira:
> @adpi2 I just rebased, fixed conflicts and re-ran `generateContrabands` and but it did create some suspicious diffs (i.e. check `ivy/src/main/contraband-scala/sbt/librarymanagement/ivy/ExternalIvyConfigurationFormats.scala`)

It is not ideal, but can you revert those diffs? I think they are mostly formatting...

mdedetrich:
> > @adpi2 I just rebased, fixed conflicts and re-ran `generateContrabands` and but it did create some suspicious diffs (i.e. check `ivy/src/main/contraband-scala/sbt/librarymanagement/ivy/ExternalIvyConfigurationFormats.scala`)
> 
> It is not ideal, but can you revert those diffs? I think they are mostly formatting...

So this is a result of running compile that runs scalafmt, should i still revert it?

As a related side note, I would suggest not running scalafmt on compile but instead to use it as a github action via scala-native-formatter. There is a PR for this open at https://github.com/sbt/librarymanagement/pull/449 but I prefer the separate github action because its faster and also doesn't break other parts of the build (let me know if you want to go ahead with this)

adpi2:
Moved to https://github.com/sbt/sbt/pull/7753

PR #384, opened 2021-07-08
Fix Scala 3 binary version

### LINKED ISSUE #6585: Incorrectly calculated binary scala versions for scala 3
## problem

`binaryScalaVersion` in librarymanagement doesn't always calculate the version correctly for scala 3, e.g. (input -> output):
`3.0.1-bin-nonbootstrapped` -> `3.0.1null`
`3.0.1-SNAPSHOT` -> `3.0.1-SNAPSHOT`
The problem seems not to occur for milestones and release candidate versions.

## expectation

The binary version of all scala versions later that `3.0.0` should be `3`

## notes

This affects the release of scala 3.0.1

adpi2:
Fixes https://github.com/sbt/sbt/issues/6585

The binary version of `3.x.y-foo-bar` should be `3` if x > 0 or  y > 0.

Example: 
- for `3.0.1-bin-nonbootstrapped` it should be `3`
- for `3.0.1-SNAPSHOT` it should be `3`

@smarter @prolativ

### DIFF
--- a/core/src/main/scala/sbt/internal/librarymanagement/cross/CrossVersionUtil.scala
+++ b/core/src/main/scala/sbt/internal/librarymanagement/cross/CrossVersionUtil.scala
@@ -76,11 +76,12 @@ object CrossVersionUtil {
     }
 
   private[sbt] def binaryScala3Version(full: String): String = full match {
-    case ReleaseV(maj, _, _, _)                                               => maj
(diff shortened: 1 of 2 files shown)


prolativ:
Not sure if we should care about versions with missing patch number like `3.1` or `3.2-bin-foo` or not. In theory this shouldn't be a problem if we never decide to release such versions

smarter:
Compiler version numbers should always have a patch field yes.

eed3si9n (approved):
Thanks!

PR #386, opened 2021-09-17
Avoid repeating versions in Eviction error message

rtyley:
Currently the evicted version numbers can be repeated many times:

> [error] 	* org.scala-lang.modules:scala-java8-compat_2.13:1.0.0 (early-semver) is selected over {0.9.0, 0.9.0, 0.9.0, 0.9.0, 0.9.1, 0.9.1, 0.9.1}

This change makes for a clearer message by removing that unnecessary noisy repetition:

> [error]         * org.scala-lang.modules:scala-java8-compat_2.13:1.0.0 (early-semver) is selected over {0.9.0, 0.9.1}

Thanks!


eed3si9n (approved):
Thanks!

PR #391, opened 2021-12-22
Respect maven.repo.local property

peter-janssen:
The local repository location is derived from the maven settings.xml. However there is also the `maven.repo.local` system property to override this ([docs](https://maven.apache.org/ref/3.8.4/maven-model-builder/index.html)). It would be nice if sbt would respect this. 
The content of this pull request is a naive approach. Happy to receive some feedback.

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
+++ b/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
@@ -392,7 +392,8 @@ private[librarymanagement] abstract class ResolverFunctions {
           System.err.println(s"WARNING: Problem parsing ${f().getAbsolutePath}, ${e.getMessage}");
           None
       }
-    loadHomeFromSettings(() => new File(sbt.io.Path.userHome, ".m2/settings.xml")) orElse
+    sys.props.get("maven.repo.local").map(new File(_)) orElse
+      loadHomeFromSettings(() => new File(sbt.io.Path.userHome, ".m2/settings.xml")) orElse
       loadHomeFromSettings(() => new File(new File(System.getenv("M2_HOME")), "conf/settings.xml")) getOrElse
       new File(sbt.io.Path.userHome, ".m2/repository")
   }


core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -397,8 +397,16 @@ private[librarymanagement] abstract class ResolverFunctions {
       new File(sbt.io.Path.userHome, ".m2/repository")
   }
   // TODO - should this just be the *exact* same as mavenLocal?  probably...
-  def publishMavenLocal: MavenCache = new MavenCache("publish-m2-local", mavenLocalDir)
-  def mavenLocal: MavenRepository = new MavenCache("Maven2 Local", mavenLocalDir)
+  def publishMavenLocal: MavenCache =
+    new MavenCache(
+      "publish-m2-local",
+      Option(System.getProperty("maven.repo.local")).map(new File(_)) getOrElse mavenLocalDir
```
eed3si9n: I'm generally fine with the idea of respecting the system property used by Maven, but I feel like this logic should go into `mavenLocalDir` just above it that goes through multiple stages like looking for `M2_HOME` environment variable to figure out the Maven local location.

peter-janssen: Agreed, somehow I was under the impression that `mavenLocalDir` was supposed to be the maven home directory.

core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -392,7 +392,8 @@ private[librarymanagement] abstract class ResolverFunctions {
           System.err.println(s"WARNING: Problem parsing ${f().getAbsolutePath}, ${e.getMessage}");
           None
       }
-    loadHomeFromSettings(() => new File(sbt.io.Path.userHome, ".m2/settings.xml")) orElse
+    Option(System.getProperty("maven.repo.local")).map(new File(_)) orElse
```
eed3si9n:
```suggestion
    sys.props.get("maven.repo.local").map(new File(_)) orElse
```

nit for Scalaism

peter-janssen: Today I learned...

eed3si9n (approved):
Thanks for the contribution!

PR #393, opened 2022-01-16
Deprecate `sonatypeRepo`, add `sonatypeOssRepos` to `Resolver`

armanbilge:
Riffing on https://github.com/sbt/librarymanagement/pull/392#issuecomment-1013844695.

Since sonatype has indicated they may/will add additional hosts, this seems like the most sensible way forward.

> we can always cap subscribership on https://s01.oss.sonatype.org at a level that ensures peak performance and spin up yet another new host to handle future load.

https://central.sonatype.org/news/20210223_new-users-on-s01/#why-are-we-doing-this

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
+++ b/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
@@ -155,7 +155,7 @@ private[librarymanagement] abstract class ResolverFunctions {
     url("sbt-plugin-" + status, new URL(SbtRepositoryRoot + "/sbt-plugin-" + status + "/"))(
       ivyStylePatterns
     )
-  @deprecated("Use sonatypeRepos instead", "1.7.0")
+  @deprecated("Use sonatypeOssRepos instead", "1.7.0")
   def sonatypeRepo(status: String) =
     MavenRepository(
       "sonatype-" + status,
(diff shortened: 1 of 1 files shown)


core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -153,12 +155,20 @@ private[librarymanagement] abstract class ResolverFunctions {
...
       if (status == "releases") SonatypeReleasesRepository
       else SonatypeRepositoryRoot + "/" + status
     )
+  private def sonatypeS01Repo(status: String) =
+    MavenRepository(
+      "sonatype-s01-" + status,
+      SonatypeS01RepositoryRoot + "/" + status
+    )
+  def sonatypeRepos(status: String) =
+    Vector(sonatypeRepo(status): @nowarn("cat=deprecation"), sonatypeS01Repo(status))
```
eed3si9n: The difference between `sonatypeRepo(...)` and `sonatypeRepos(...)` is a bit subtle, so I'd suggest `allSonatypeOssRepos(...)`.

armanbilge: Good point. I like `sonatypeOssRepos`: starting with `sonatype` is good for discoverability, and they are all `*oss.sonatype.org` so I suppose there could be other non-OSS repos.

armanbilge:
@eed3si9n do you think this one might land in sbt 1.7.0? `org.typelevel`, `org.http4s`, and `org.tpolecat` have migrated to s01 so there is a real need for this. Thanks! :)

eed3si9n (approved):

eed3si9n:
@armanbilge yea. 🚢 it.

PR #396, opened 2022-02-03
Add commonly used values to License

marconilanna:


eed3si9n:
Thanks for the contribution. I appreciate that you took time to make this change, but I feel like there's a value in having fewer licenses in the predefined list. After all, current `build.sbt` supports freeform `(String, URL)`.
For any new projects considering an open source license, I don't think any of those choices add much value, but become more of a confusion since CC0 > MIT > Apache 2.0 > GPL v3 represents clear progression of clauses/restriction.
Also the data shows 75% of the licenses are Apache 2.0 + MIT.

PR #409, opened 2022-12-21
Try resolve sbt plugin from valid Maven pattern

### LINKED ISSUE #3410: The maven pattern for sbt plugins is invalid based on the spec
I've been in touch with the Bintray team because of https://github.com/sbt/sbt-bintray/pull/125.
  
It happens that the reason why sbt bintray does not accept the fix in the aforementioned PR is because **sbt has been using an incorrect Maven pattern for sbt plugins**. This is the reply I got from the Bintray team:
  
> Going off of the Maven Design Documentation, Bintray expects the following Maven layout format:
`[orgPath]/[module]/[baseRev](-[folderItegRev])/[module]-[baseRev](-[fileItegRev])(-[classifier]).[ext]`. Note the “module” field appears twice.
  
Our pattern is the following: https://github.com/sbt/librarymanagement/blob/53c80f076a5e9ab469037f352a4d68173140e216/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L357-L358, which is obviously wrong (there's a concept of module and a concept of artifact, where both should be the same).
  
The reason why this works for Sonatype is because they have a permissive layout for projects whose publishing rules do not abide by the spec. Bintray doesn't, and that's why no one has succeeded to publish sbt plugins to Bintray maven repositories.
  
This is a major issue that IMO should be sorted out as soon as possible. The only way I see to solve this problem is to add another pattern to sbt and make it use alongside the wrong pattern (we cannot remove it, too many projects depend on it).

adpi2:
This PR is part of https://github.com/sbt/sbt/pull/7096 whose goal is to fix https://github.com/sbt/sbt/issues/3410 in a smooth way. For a full description please read https://github.com/sbt/sbt/pull/7096.

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/ConvertResolver.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/ConvertResolver.scala
@@ -27,10 +27,16 @@ import org.apache.ivy.plugins.resolver.{
 import org.apache.ivy.plugins.repository.url.{ URLRepository => URLRepo }
 import org.apache.ivy.plugins.repository.file.{ FileResource, FileRepository => FileRepo }
 import java.io.{ File, IOException }
+import java.util.Date
 
-import org.apache.ivy.util.{ ChecksumHelper, FileUtil, Message }
 import org.apache.ivy.core.module.descriptor.{ Artifact => IArtifact }
+import org.apache.ivy.core.module.id.ModuleRevisionId
+import org.apache.ivy.core.module.descriptor.DefaultArtifact
 import org.apache.ivy.core.report.DownloadReport
+import org.apache.ivy.plugins.resolver.util.{ ResolvedResource, ResourceMDParser }
+import org.apache.ivy.util.{ ChecksumHelper, FileUtil, Message }
+import scala.collection.JavaConverters._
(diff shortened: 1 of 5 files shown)


julienrf (commented):
That looks great! Is it possible to test the resolution of sbt plugins?

eed3si9n:
@adpi2  Overall looks good to me as well. As I understood the conversation on Slack, the magic is that POM's `<dependency>` entry is transformed based on the extra attribute information available in Maven properties in the case of sbt plugins. For future reference, could you add a note to that effect in the commit message please?

adpi2:
I added a few more comments, and I rewrote the commit message.

Also I suggest we wait a bit before merging, until we agree on the implementation in https://github.com/coursier/coursier/pull/2633.

julienrf (approved):

adpi2:
Coursier is now ready to resolve the new Maven pattern of sbt plugins:
- https://github.com/coursier/coursier/pull/2633
- https://github.com/coursier/sbt-coursier/pull/445
- https://github.com/coursier/sbt-coursier/releases/tag/v2.0.16

@eed3si9n If you agree, I need your :heavy_check_mark: to merge this PR.

eed3si9n (approved):
Thanks @adpi2!

PR #411, opened 2023-02-18
Add Apache Maven Snapshots Repository

mdedetrich:
This PR adds the Apache Maven Snapshots repository (see https://infra.apache.org/repository-faq.html#basic for more details). Note that the releases snapshot repo does not need to be added since its already synced to Maven central, but this should be helpful for sbt users that want to use Apache software snapshots.

Afaik, the `name` field for Maven Repository seems to be purely illustrative, so I just used `apache-snapshots` which seems to follow sbt convention (can also do apache-maven-snapshots if necessary).


eed3si9n (approved):
Thanks!

PR #412, opened 2023-03-26
Add plugin-tool configuration

eed3si9n:
This adds plugin-tool configuration as a sandbox configuration for plugins.

See https://eed3si9n.com/sbt-drop-custom-config/ for details.


sjrd:
As I explain in https://github.com/sbt/sbt/discussions/7189#discussioncomment-5442563, this will not help, because it will not allow *two* plugins to use that unique configuration for their independent purposes. So no plugin conscious about not stepping on another plugin's toes is going to use `plugin-tool`.

eed3si9n:
Closing this for now.

PR #413, opened 2023-04-09
Expand properties found in maven settings

### LINKED ISSUE #7195: Non-compliant parsing of maven's settings.xml file
## steps

Have a non-default maven [`settings.xml`](https://maven.apache.org/settings.html) file on your system, in which you customized the `localRepository` element to point to an environment variable.

This is perfectly legal and is, in fact, the first example in the [Simple values](https://maven.apache.org/settings.html#simple-values) section:
```xml
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0 https://maven.apache.org/xsd/settings-1.0.0.xsd">
  <localRepository>${user.home}/.m2/repository</localRepository>
...
```

Then attempt to run `sbt publishM2` in some valid project.

## problem

The variable, instead of being expanded, will be used as is in the path, resulting in artifacts being published to, say, `./${env.FOO_BAR}/path/to/artifacts`.

## expectation

Variables should be expended, as documented in the `settings.xml` documentation.

## notes

Here's the [source of the problem](https://github.com/sbt/librarymanagement/blob/develop/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala#L400-L402):
```scala
          ((XML.loadFile(file) \ "localRepository").text match {
            case ""    => None
            case e @ _ => Some(new File(e))
```

Notice how no expansion is being performed in the content of the node.

nrinaudo:
Fixes sbt/sbt#7195

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
+++ b/core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
@@ -391,6 +391,17 @@ private[librarymanagement] abstract class ResolverFunctions {
   def defaultRetrievePattern =
     "[type]s/[organisation]/[module]/" + PluginPattern + "[artifact](-[revision])(-[classifier]).[ext]"
   final val PluginPattern = "(scala_[scalaVersion]/)(sbt_[sbtVersion]/)"
+  private[librarymanagement] def expandMavenSettings(str: String): String = {
+    // Aren't regular expressions beautifully clear and concise.
+    // This means "find all ${...}" blocks, with the first group of each being the text between curly brackets.
+    val findQuoted = "\\$\\{([^\\}]*)\\}".r
+    val env = "env\\.(.*)".r
+
+    findQuoted.replaceAllIn(str, _.group(1) match {
+      case env(variable) => sys.env.getOrElse(variable, "")
+      case property      => sys.props.getOrElse(property, "")
+    })
+  }
   private[this] def mavenLocalDir: File = {
     def loadHomeFromSettings(f: () => File): Option[File] =
       try {
@@ -399,7 +410,7 @@ private[librarymanagement] abstract class ResolverFunctions {
         else
           ((XML.loadFile(file) \ "localRepository").text match {
             case ""    => None
-            case e @ _ => Some(new File(e))
+            case e @ _ => Some(new File(expandMavenSettings(e)))
           })
       } catch {
         // Occurs inside File constructor when property or environment variable does not exist
--- a/core/src/test/scala/sbt/librarymanagement/ResolverExtraTest.scala
+++ b/core/src/test/scala/sbt/librarymanagement/ResolverExtraTest.scala
@@ -1,0 +1,43 @@
+package sbt.librarymanagement
+
+import verify.BasicTestSuite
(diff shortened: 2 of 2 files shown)


core/src/main/scala/sbt/librarymanagement/ResolverExtra.scala
```diff
@@ -391,6 +391,17 @@ private[librarymanagement] abstract class ResolverFunctions {
   def defaultRetrievePattern =
     "[type]s/[organisation]/[module]/" + PluginPattern + "[artifact](-[revision])(-[classifier]).[ext]"
   final val PluginPattern = "(scala_[scalaVersion]/)(sbt_[sbtVersion]/)"
+  private[librarymanagement] def expandMavenSettings(str: String): String = {
```
nrinaudo, line 394: Rather than scoping that to `mavenLocalDir`, I exposed it for testing purposes. Happy to backpedal on that, it doesn't seem to be the usual coding style of this project, but I felt tests would be useful for anything that involves regular expressions.

core/src/test/scala/sbt/librarymanagement/ResolverExtraTest.scala
```diff
@@ -0,0 +1,43 @@
+package sbt.librarymanagement
+
+import verify.BasicTestSuite
+import scala.annotation.nowarn
+
+@nowarn // Necessary because our test cases look like interpolated strings.
+object ResolverExtraTest extends BasicTestSuite {
```
nrinaudo, line 7: One weirdness that I haven't had the courage to look into: this was initially a `UnitSpec`, like `ResolverTest`, but tests didn't get executed.

Not sure whether this is a bug or working as intended, so I didn't change it. Note that `ResolverTest` isn't getting executed, at least on my local setup.

It would probably be good to either document or fix this, depending on whether it's desired or a bug, as it's quite easy to write tests that fail but aren't executed, and think your tests pass. Happy to help with this once the situation is clarified.

nrinaudo:
As a first time contributor, I was rather confused at first - `semanticdb` wouldn't resolve, and I had to pretend to be running my builds in a CI environment via the following in `build.sbt`:
```scala
Global / semanticdbEnabled := false
```

This seems either undesirable or in need of documentation, as it adds unnecessary friction to the experience.

eed3si9n (approved):
Looks great. Thanks for the contribution!

core/src/test/scala/sbt/librarymanagement/ResolverExtraTest.scala
eed3si9n: Not sure what it could be. Maybe it's due the fact that some are ScalaTest that uses class and others are verify that uses object?

nrinaudo: Honestly not a clue. There's a `verify` runner explicitly registered in the build, which I initially thought might override existing ones, but I see that *some* ScalaTest tests are being executed so... very confused.

PR #415, opened 2023-05-04
libraryDependencySchemes overrides assumedVersionScheme

### LINKED ISSUE #6745: libraryDependencySchemes doesn't seem to work for scala-java8-compat
## steps

Reproducible example on Scastie: https://scastie.scala-lang.org/hRyPIl7IQiCLzo2RXG6gwg

Using SBT 1.5.6. Not necessarily restricted to 1.5.6, I haven't checked other 1.5.x. I'm not sure what version of SBT scastie is using, but it fails the same as 1.5.6 in a local project.

With following dependencies:

- akka-actor 2.6.0, which depends on scala-java8-compat 0.8.0.
- swagger-akka-http 1.4.0, which depends on scala-java8-compat 0.9.1.

With following eviction settings:

```
assumedEvictionErrorLevel := Level.Error
assumedVersionScheme := VersionScheme.EarlySemVer
libraryDependencySchemes += "org.scala-lang.modules" %% "scala-java8-compat" % VersionScheme.Always
```

## problem

`sbt evicted` still returns an error:

```
java.lang.RuntimeException: found version conflict(s) in library dependencies; some are suspected to be binary incompatible:

	* org.scala-lang.modules:scala-java8-compat_2.12:0.9.1 (early-semver?) is selected over 0.8.0
	    +- com.github.swagger-akka-http:swagger-akka-http_2.12:1.4.0 (depends on 0.9.1)
	    +- com.typesafe.akka:akka-actor_2.12:2.6.0            (depends on 0.8.0)


this can be overridden using libraryDependencySchemes or evictionErrorLevel
	at scala.sys.package$.error(package.scala:30)
	at sbt.internal.LibraryManagement$.resolve$1(LibraryManagement.scala:89)
	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$12(LibraryManagement.scala:133)
	at sbt.util.Tracked$.$anonfun$lastOutput$1(Tracked.scala:73)
	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$20(LibraryManagement.scala:146)
	at scala.util.control.Exception$Catch.apply(Exception.scala:228)
	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11(LibraryManagement.scala:146)
	at sbt.internal.LibraryManagement$.$anonfun$cachedUpdate$11$adapted(LibraryManagement.scala:127)
	at sbt.util.Tracked$.$anonfun$inputChangedW$1(Tracked.scala:219)
	at sbt.internal.LibraryManagement$.cachedUpdate(LibraryManagement.scala:160)
	at sbt.Classpaths$.$anonfun$updateTask0$1(Defaults.scala:3678)
	at scala.Function1.$anonfun$compose$1(Function1.scala:49)
	at sbt.internal.util.$tilde$greater.$anonfun$$u2219$1(TypeFunctions.scala:62)
	at sbt.std.Transform$$anon$4.work(Transform.scala:68)
	at sbt.Execute.$anonfun$submit$2(Execute.scala:282)
	at sbt.internal.util.ErrorHandling$.wideConvert(ErrorHandling.scala:23)
	at sbt.Execute.work(Execute.scala:291)
	at sbt.Execute.$anonfun$submit$1(Execute.scala:282)
	at sbt.ConcurrentRestrictions$$anon$4.$anonfun$submitValid$1(ConcurrentRestrictions.scala:265)
	at sbt.CompletionService$$anon$2.call(CompletionService.scala:64)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1135)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:831)
```

## expectation

From my understanding, the `libraryDependencySchemes` setting to `Always` should mean that versions of scala-java8-compat library are _always_ compatible. So the 0.8.0 and 0.9.1 incompatibility should be ignored, and the error should not be thrown.

## notes

Is setting the version scheme for a specific library to "always" the correct way to specify "don't check the version compatibility of this library"? That's what it seems like from the docs and release notes, but maybe I've missed someting.

adriaanm:
Allow user to suppress eviction errors for a specific library.

We want the user to be able to suppress eviction errors for a specific library,
even (especially?) if they would result in an incompatible eviction
based on the assumed version scheme.

So, only fall back to the assumed scheme if there is no given scheme by the user or the pom.

Fix https://github.com/sbt/sbt/issues/6745

### DIFF
--- a/core/src/main/scala/sbt/librarymanagement/EvictionError.scala
+++ b/core/src/main/scala/sbt/librarymanagement/EvictionError.scala
@@ -88,33 +88,34 @@ object EvictionError {
       // https://github.com/sbt/sbt/issues/4946
       case p if p.winner.isDefined =>
         val winner = p.winner.get
-        def lookupUserDefined(org: String = "*", mod: String = "*") =
+        def fromLibraryDependencySchemes(org: String = "*", mod: String = "*") =
           userDefinedSchemes.get((org, mod))
         def fromWinnerPom = VersionSchemes.extractFromExtraAttributes(
           winner.extraAttributes.toMap ++ winner.module.extraAttributes
         )
         // prioritize user-defined version scheme to allow overriding the real scheme
         val userDefinedSchemeOrFromPom =
-          lookupUserDefined(p.organization, p.name)
-            .orElse(lookupUserDefined(p.organization))
+          fromLibraryDependencySchemes(p.organization, p.name)
+            .orElse(fromLibraryDependencySchemes(p.organization))
             .orElse(fromWinnerPom)
-            .orElse(lookupUserDefined())
+            .orElse(fromLibraryDependencySchemes())
 
         val assumedScheme =
           if (isNameScalaSuffixed(p.name)) assumedVersionScheme
           else assumedVersionSchemeJava
 
-        def hasIncompatibleVersionForScheme(scheme: Option[String]) = {
+        def hasIncompatibleVersionForScheme(scheme: String) = {
           val isCompat =
-            scheme.map(VersionSchemes.evalFunc).getOrElse(EvictionWarningOptions.guessTrue)
+            VersionSchemes.evalFunc(scheme)
           p.evicteds.exists { r =>
             !isCompat((r.module, Some(winner.module), module.scalaModuleInfo))
           }
         }
 
-        if (hasIncompatibleVersionForScheme(userDefinedSchemeOrFromPom))
+        val userDefinedSchemeOrAlways = userDefinedSchemeOrFromPom.getOrElse(VersionSchemes.Always)
+        if (hasIncompatibleVersionForScheme(userDefinedSchemeOrAlways))
           incompatibleEvictions += (p -> userDefinedSchemeOrFromPom.getOrElse("?"))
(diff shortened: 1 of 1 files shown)


eed3si9n:
Overall LGTM. Thanks for the contribution!

I think I did add some (integrationish) test in https://github.com/sbt/librarymanagement/blob/develop/ivy/src/test/scala/sbt/internal/librarymanagement/EvictionErrorSpec.scala when I first added EvictionError. <s>Could you kindly repro the original bug and add test for it using scala-java8-compat please?</s>

https://github.com/sbt/librarymanagement/pull/416

adriaanm:
Thanks for merging! Sorry, missed the notification! I've gone dark on github for some time and forgot to re-enable them.

PR #418, opened 2023-06-25
avoid deprecated `java.net.URL` constructor

xuwei-k:
- https://bugs.openjdk.org/browse/JDK-8295949
- openjdk/jdk@4338f52


eed3si9n (approved):
Thanks!

PR #425, opened 2023-08-26
Alias withLatestSnapshots to withCachedSnapshots

mdedetrich:
The main reason behind this change is the current naming isn't really clear (even as a native english speaker) and its already caused confusion (see https://github.com/apache/incubator-pekko-connectors/pull/239#discussion_r1305807641 as an example).

The suggested name change at least in my view makes a lot more clear whats going on, i.e. sbt will cache snapshots versions so they don't re-resolve. I am not too fussed about this name specifically, there are others as well i.e.

* `withAlwaysRefreshSnasphots`
* `withAlwaysCheckSnasphots`

If this PR gets merged I will also update the core sbt build/documentation to reflect this

### DIFF
--- a/ivy/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
+++ b/ivy/src/main/scala/sbt/internal/librarymanagement/Ivy.scala
@@ -541,7 +541,7 @@ private[sbt] object IvySbt {
       val delegate = ivyint.SbtChainResolver(delegatedName, rest, settings, updateOptions, log)
       val initialResolvers = projectResolvers :+ delegate
       val freshOptions = UpdateOptions()
-        .withLatestSnapshots(false)
+        .withCachedSnapshots(true)
         .withModuleResolvers(updateOptions.moduleResolvers)
       ivyint.SbtChainResolver(name, initialResolvers, settings, freshOptions, log)
     }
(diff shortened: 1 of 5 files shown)


mdedetrich:
I have just updated the PR so that it compiles properly (there were some basic issues beforehand).

One thing to note is that I did not reliaze that we serialize `UpdateOptions` to JSON. The PR handles this so that we don't have regressions however I left a note indicating that we might be able to remove the workarounds in sbt 2.x.x (i.e. if sbt 2.x.x happens to serliaize this data in another folder, not sure if this is the case or not).

eed3si9n:
It's not the opposite of `withCachedSnapshots`, but rather it's the opposite of `withFirstSnapshotFound`.

According to the release note section that likely I wrote in https://github.com/sbt/sbt/releases/tag/v0.13.6

> sbt 0.13.6 adds a new setting key called `updateOptions` for customizing the details of managed dependency resolution with `update` task. One of its flags is called `lastestSnapshots`, which controls the behavior of the chained resolver. Up until 0.13.6, sbt was picking the first `-SNAPSHOT` revision it found along the chain. When `latestSnapshots` is enabled (default: `true`), it will look into all resolvers on the chain, and compare them using the publish date.
>
> The tradeoff is probably a longer resolution time if you have many remote repositories on the build or you live away from the severs. So here's how to disable it:

```scala
updateOptions := updateOptions.value.withLatestSnapshots(false)
```

Also I think this only works for Ivy.

PR #428, opened 2023-09-10
Add allModuleReports to UpdateReport

mdedetrich:
I came across this when doing an upcoming change in sbt-license-report (specifically I want to remove sbt-license-report manual usage of Ivy and replace it with `DependencyResolution`/`transitiveUpdate` both of which return an `UpdateReport`). 

Hence this method is needed in order to retrieve extra information (such as license, homepage, etc etc) from an `UpdateReport`


eed3si9n (approved):
Thanks!

PR #431, opened 2023-09-11
Update github checkout action

mdedetrich:
Minor change but a new version of the github checkout action was released


eed3si9n (approved):
Thanks!

PR #433, opened 2023-09-16
Avoid repeating versions in Eviction warning message

rtyley:
As with PR https://github.com/sbt/librarymanagement/pull/386, which dealt with Eviction **errors**, this fixes the way `EvictionWarning` reports the list of evicted versions - removing duplicate revisions that look like this:

```
* org.scala-lang.modules:scala-java8-compat_2.13:1.0.2 is selected over {1.0.0, 1.0.0}
```

This fix does a refactor so that both `EvictionError` & `EvictionWarning` are using the same logic to generate the revision string - the logic is moved to the new field `EvictionPair.evictedRevs`.


eed3si9n (approved):
Thanks!

PR #435, opened 2023-10-23
set overrideScalaVersion to false by default

lrytz:
One of the pieces needed for [SIP-51](https://docs.scala-lang.org/sips/drop-stdlib-forwards-bin-compat.html)


lrytz:
Ah this needs to be narrowed down to Scala 2.13 projects, so I guess we need to keep the defaults here and change the setting in sbt according to the Scala version.

PR #436, opened 2024-03-30
Update jsch to 0.2.17 (switch to the com.github.mwiede fork)

azolotko:


eed3si9n:
@azolotko Thanks for the contribution! Are there specific use case you're trying to unblock by this switch to Jsch fork?

azolotko:
> @azolotko Thanks for the contribution! Are there specific use case you're trying to unblock by this switch to Jsch fork?

Yes, I’d like to use EdDSA/Ed25519 keys in my sbt-based task automation. The above mentioned fork [does support them](https://github.com/mwiede/jsch?tab=readme-ov-file#are-ssh-ed25519-ssh-ed448-curve25519-sha256-curve448-sha512--chacha20-poly1305opensshcom-supported).

One thing to mention is that I’m not using the jsch from sbt directly, but rather via https://github.com/dacr/jassh, of which jsch is a dependency. Simply overriding its dependency on jsch doesn’t help because sbt has its own jsch on the classpath.

eed3si9n (approved):

PR #444, opened 2024-06-28
Preserve backslashes in `expandMavenSettings`

desbo:
We use `expandMavenSettings` to turn environment variable expressions in `settings.xml` into their actual values.

Before this change, if an environment variable value contained backslashes, they would have been treated as escape characters by `Regex.replaceAllIn` and removed from the returned string. This caused an issue when using backslashes for Windows paths.


eed3si9n (approved):
@desbo Thanks for the contribution!

PR #457, opened 2024-10-18
Replace scalafmtOnCompile with CI check

mdedetrich:
This PR replaces `scalafmtOnCompile` with a github action that runs scalafmt-native that checks if the entire codebase is formatted. The advantage of this versus using `scalafmtOnCompile` is it doesn't mess up editors while developing and since it is its own github action check it runs in parallel to the main sbt build (this also means it will even run if the sbt fails to load/compile). The scalafmt-native check also only runs on files that are different to `origin/develop` and since its running in native the entire github check typically only takes around 5-10 seconds.

Most critically it ensures that the code is actually properly formatted since with the current setup, its still possible to push unformatted code just by simply forgetting to compile before pushing (which is what happened before).

The only thing that needs to be done after this PR is merged is to add `Scalafmt / Code is formatted` as a github CI status check so that you cannot merge PR's if they are not formatted.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,23 @@
+name: Scalafmt
+
+permissions: {}
+
+on:
+  pull_request:
+    branches: ['**']
+
+jobs:
+  build:
+    name: Code is formatted
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout current branch (full)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Check project is formatted
+        uses: jrouly/scalafmt-native-action@v3
+        with:
+          arguments: '--list'
--- a/.scalafmt.conf
+++ b/.scalafmt.conf
@@ -1,4 +1,4 @@
-version = 3.7.4
+version = 3.8.3
 maxColumn = 100
 project.git = true
 project.excludeFilters = [ /sbt-test/, /input_sources/, /contraband-scala/ ]
--- a/build.sbt
+++ b/build.sbt
@@ -26,7 +26,6 @@ ThisBuild / scmInfo := {
   Some(ScmInfo(url(s"https://github.com/$slug"), s"git@github.com:$slug.git"))
 }
 ThisBuild / licenses := List(("Apache-2.0", url("https://www.apache.org/licenses/LICENSE-2.0")))
-ThisBuild / scalafmtOnCompile := true
 ThisBuild / developers := List(
   Developer("harrah", "Mark Harrah", "@harrah", url("https://github.com/harrah")),
   Developer("eed3si9n", "Eugene Yokota", "@eed3si9n", url("http://eed3si9n.com/")),
@@ -67,8 +66,6 @@ def commonSettings: Seq[Setting[_]] = Def.settings(
   inCompileAndTest(
     (console / scalacOptions) --= Vector("-Ywarn-unused-import", "-Ywarn-unused", "-Xlint")
   ),
-  scalafmtOnCompile := true,
-  Test / scalafmtOnCompile := true,
   Compile / publishArtifact := true,
   Test / publishArtifact := false,
   Test / parallelExecution := false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,4 +40,4 @@ jobs:
       if: ${{ matrix.jobtype == 1 }}
       shell: bash
       run: |
-        sbt -v -Dfile.encoding=UTF8 scalafmtCheckAll +test +packagedArtifacts
+        sbt -v -Dfile.encoding=UTF8 +test +packagedArtifacts


Friendseeker:
This looks fantastic. I am not too sure about disabling `scalafmtOnCompile` (I use Intellij and `scalafmtOnCompile` never caused any issue) but adding a CI job to check formatting is certainly very welcomed.

Friendseeker:
Just want to confirm, does the new CI job also catch an ill-formatted `build.sbt` file? If so then the PR LGTM.

Friendseeker (approved):

Friendseeker:
Wait looks like there's already a `scalafmt` check as part of CI. If it does the same thing as the PR, I guess the PR is no longer necessary then.

https://github.com/sbt/librarymanagement/blob/55893d357bcc5a0cdb617ba1cf24491023839cf6/.github/workflows/ci.yml#L39-L43

---

Ehh Github does not have a button to rollback PR approval...

Friendseeker (changes requested):
Due to https://github.com/sbt/librarymanagement/pull/457#issuecomment-2452637422, reserving approval for now.

mdedetrich:
> Wait looks like there's already a `scalafmt` check as part of CI. If it does the same thing as the PR, I guess the PR is no longer necessary then.

Let me change it, sec

mdedetrich:
@Friendseeker Done

Friendseeker:
> @Friendseeker Done

Thanks! I am not very familiar with the repo as evidenced by me initially overlooking the existence of 'scalafmtcheckAll'. Probably for the best to let Eugene or other people more experienced with the repo to proceed with the review.

build.sbt
```diff
@@ -26,7 +26,6 @@ ThisBuild / scmInfo := {
   Some(ScmInfo(url(s"https://github.com/$slug"), s"git@github.com:$slug.git"))
 }
 ThisBuild / licenses := List(("Apache-2.0", url("https://www.apache.org/licenses/LICENSE-2.0")))
-ThisBuild / scalafmtOnCompile := true
```
eed3si9n, line 29: The reason why we turn this setting on is because when someone sends a PR, half the time they won't remember to run scalafmt. Maybe Scalafmt has been popularized enough that this is no longer an issue? I wish we could just append a format commit, which is what we do at work.

mdedetrich: > The reason why we turn this setting on is because when someone sends a PR, half the time they won't remember to run scalafmt. Maybe Scalafmt has been popularized enough that this is no longer an issue? I wish we could just append a format commit, which is what we do at work.

On the flip side, if someone does code changes and doesn't run compile then they can make a PR with unformatted code which can get merged (this actually happened with librarymanagement).

The intention here is that even if someone submits an unformatted PR, the PR won't actually be able to be merged (on the assumption you add it as a ci check).

Over the years I have tried many different approaches and find this one to be the most suitable for everyone's work style and actually enforced that the code base stays formatted which is the whole point.

eed3si9n: We run `scalafmtCheckAll` on CI, so I'm not sure how unformatted code got slipped? Maybe we don't enforce linear history so scalafmt version change didn't catch some PRs?

mdedetrich: I don't know how it happened but when we were doing the sbt 2.x breaking changes the codebase was not in a formatted state (which was quite annoying when working on it)

In any case, the setup this is suggested in this PR is the same as many sbt plugins and other OS projects as well (I have contributed this setup to many projects) so it shouldn't be too controversial.

.github/workflows/format.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: Checkout current branch (full)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          persist-credentials: false
+
+      - name: Check project is formatted
+        uses: jrouly/scalafmt-native-action@v3
+        with:
+          arguments: '--list --mode diff-ref=origin/develop'
```
eed3si9n: Wouldn't this diff-ref delay the formatting of the code when Scalafmt version is changed, and end up creating spurious changes when the next person comes along to touch a Scala file?

.scalafmt.conf
```diff
@@ -1,4 +1,4 @@
-version = 3.7.4
+version = 3.8.3
```
eed3si9n, line 1: For example, we have no idea what this bump would do to the code?

mdedetrich: The codebase was already formatted before the version bump and since there is no diff in the code it means there was no breaking changes.

I bumped the version because for some reason the current version of scalafmt cli didn't run on my laptop with the project.git and/or project.layout option, the newest version does (likely some big fix)

eed3si9n: Since it's going to run in CI machine with its own job, I'd suggest dropping `diff-ref` and check all the files.

mdedetrich: > Since it's going to run in CI machine with its own job, I'd suggest dropping `diff-ref` and check all the files.

Sure, diff-ref makes a difference in large codebases but this project is not in that state.

I'll do the change when I get home

mdedetrich: @eed3si9n This is now done

eed3si9n (approved):
Thanks!

PR #492, opened 2025-02-18
Upgrade sbt ivy

mkurz:


eed3si9n:
lm is part of sbt/sbt in develop.

